PR #33, opened 2015-02-26
Add a differently structured version

quis:
This pull request:

## Restructures the existing content into a new folder (`g6/`) which contains:
- A manifest called `index.yml` which defines the order of pages (this would eventually be owned by the implementing app)
- One file per question (eg `elasticCloud.yml`), named after the key that is used by the data API, service page in the Grails app, admin front end, etc.

This will make it much easier to grab the relevant content for a specific question, without having to do any preprocessing of the content to extract all the questions.

## Changes validations to be an array of key/value objects, just for documents at the moment

eg
```
validations:
  - name: name_of_validation_rule
    message "Message to be shown if validation fails"
  - name: name_of_validation_rule
    message "Message to be shown if validation fails"
```

## Adds `optional` as a new key

This enables correct validation logic around questions like _SFIA rate card_.

## Removes the previous structure


robyoung:
I might not be understanding this change but if this is just restructuring existing data rather than modifying it shouldn't it be removing stuff as well as adding it?

quis:
@robyoung There are some apps (prototype, SSP) which depend on the old data. Think this branch should just stay as a branch while I'm using it for prototype-y/spike stuff.

TheDoubleK:
Deleting the old format stuff would completely break the SSP - which at this stage isn't the end of the world - but nice to not break it just yet I think.

Maybe "restructure" is the wrong word - it's more "add a differently structured version".

TheDoubleK:
Reviewing again... I may be some time...

PR #35, opened 2015-03-20
Flatten fields object

quis:
This commit does two things:

1. Put the `type` property at the top level, so that each question has a type  (eg `checkboxes`, `textarea`)
2. Rename `fields`, for questions that have multiple options, to `options`

This new format already existed for questions used by the admin app and is now consistent across all questions.

### DIFF
--- a/g6/auditInformationProvided.yml
+++ b/g6/auditInformationProvided.yml
@@ -3,16 +3,14 @@ filters: 'Consumers are aware of the audit information that will be provided to 
 hint: 'Do you provide audit information to consumers? Choose 1.'
 dependsOnLots: 'SaaS, PaaS, IaaS'
 assuranceApproach: 2answers-type1
-fields:
+type: radios
+options:
     -
-        type: radio
         label: None
     -
-        type: radio
         label: 'Data made available'
         filterLabel: 'Data made available'
     -
-        type: radio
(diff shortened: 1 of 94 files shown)


TheDoubleK:
94 files changed! I'll take a look now...

g6/auditInformationProvided.yml
```diff
@@ -3,16 +3,14 @@ filters: 'Consumers are aware of the audit information that will be provided to
 hint: 'Do you provide audit information to consumers? Choose 1.'
 dependsOnLots: 'SaaS, PaaS, IaaS'
 assuranceApproach: 2answers-type1
-fields:
+type: radios
+options:
     -
-        type: radio
         label: None
```
TheDoubleK: Does this need a 'filterLabel' too, or is 'label' is used when 'filterLabel' doesn't exist?  And if that's the case why do we have duplication for other options here? It looks like 'label' and 'filterLabel' have the same value for many questions.

TheDoubleK: Sorry, Sorry - i remember now.  FilterLabel indicates if users can filter on these options.  Sorry, as you were.

TheDoubleK:
This format works much better - especially for the 'checkboxes' and 'radios' :+1:

PR #36, opened 2015-03-23
Change validation structure

quis:
This commit applies the changes to validation structure made in #33 and applies them across all the content.


TheDoubleK:
Just thinking about this... is it right that validations are so tightly bound to the question content?  I'm imagining a situation where we might want to ask the same question in two different frameworks, but in one it's required and the other not.  Or file upload questions that have different file size limits.

So maybe the validation rules should be defined separately from the question content.

Not necessarily for now, but something to think about.

TheDoubleK:
It also feels like 'This question requires an answer.' should just be somewhere once, as a default message for that type of validation.  But again, not for now.

PR #39, opened 2015-05-18
Add basic tests

robyoung:
- Add basic tests that validate all files as YAML
- Fix index.yml to be valid yaml
- Run tests in travis


quis:
Lovely work

PR #43, opened 2015-06-16
Be explicit that "on" is a string

quis:
pyyaml 'helpfully' converts this
``` yaml
-
  on: Lot
```
to this
``` python
[
  {
    True: "Lot"
  }
]
```

By wrapping `on` in `"`quotes`"` it forces pyyaml to interpret it as a string, not a boolean, which is our intention.


allait:
Yay YAML

PR #45, opened 2015-06-18
Add content for the 'choose a lot' question

quis:
This comes from the previous SSP. Using 'description' for the text underneath each button matches the naming on http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/selection-buttons.html

### DIFF
--- a/g6/lot.yml
+++ b/g6/lot.yml
@@ -1,2 +1,32 @@
-question: 'Lot'
-type: checkboxes
+question: 'Service type'
+type: radios
+options:
+  -
+    label: Infrastructure as a Service (IaaS)
+    value: IaaS
+    description: >
+      Infrastructure is the hardware that makes software work. It’s the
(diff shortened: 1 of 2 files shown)


g6/lot.yml
```diff
@@ -1,2 +1,32 @@
-question: 'Lot'
-type: checkboxes
+question: 'Service type'
+type: radios
+options:
+  -
+    label: Infrastructure as a Service (IaaS)
+    value: IaaS
+    description: >
```
pcraig3: What does the closing angle bracket mean?

pcraig3:
Once merged, the version needs incrementing, right?

quis:
There's no versioning of the content (yet?) so you just have to [reference the SHA in bower.json](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/bower.json#L10).

PR #50, opened 2015-08-04
Make price errors more specific

robyoung:
Make the messages for min or max prices not being numbers identify the
field that is in error.

### DIFF
--- a/g6/priceString.yml
+++ b/g6/priceString.yml
@@ -16,10 +16,10 @@ validations:
     message: 'Pricing unit requires an answer. If none of the provided units apply, please choose ''Unit''.'
   -
     name: min_price_not_a_number
-    message: 'Prices must be numbers only, without units, eg 99.95'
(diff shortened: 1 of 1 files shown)


g6/priceString.yml
```diff
@@ -16,10 +16,10 @@ validations:
     message: 'Pricing unit requires an answer. If none of the provided units apply, please choose ''Unit''.'
   -
     name: min_price_not_a_number
-    message: 'Prices must be numbers only, without units, eg 99.95'
+    message: 'Minimum price must be a number, without units, eg 99.95'
   -
     name: max_price_not_a_number
-    message: 'Prices must be numbers only, without units, eg 99.95'
+    message: 'Maximum must be a number, without units, eg 99.95'
```
TheDoubleK: Why just "Maximum must..." here but "Minimum price must.." above?  Might be better with "Maximum price must..." too.

robyoung: Yes, that's a mistake, one sec.

PR #51, opened 2015-08-04
Change some questions from radios to checkboxes

TheDoubleK:
This is in line with the changes agreed with Tony R prior to the G6 SSP data export.


quis:
:+1:

PR #52, opened 2015-08-06
Make t&c question the same style as other document questions

TheDoubleK:
Currently the question for T&C document is a different style from other document upload questions - i.e:
screen shot 2015-08-06 at 15 17 47

This should make things look better.


quis:
:+1:

PR #53, opened 2015-08-10
Add validation messages for G7 declaration

robyoung:
This adds validation messages for the G7 supplier declaration form.

In addition to these error messages there is a default `answer_required` message for all non-optional fields set in the app. This is currently "This question is required".

cc/ @superroz


robyoung:
The question numbers are not correct any more.

superroz:
Could we change the validation message from 'This question is required.' to 'You must answer this question.'?

TheDoubleK:
Fixed up now.

PR #56, opened 2015-08-12
Ccs edits 120815

superroz:
a few suggested edits from CCS.


g7_declaration/PR4.yml
```diff
@@ -1,3 +1,3 @@
 question: Do you understand that you must log in to your Digital Marketplace supplier account to ask questions about the G-Cloud 7 application process?
-hint: All communication with the Crown Commercial Service (CCS) about this bid should take place through the Digital Marketplace.
+hint: All communication with the Crown Commercial Service (CCS) about this bid should take place through the Digital Marketplace. CCS will not respond to questions invididually. All answers will be published on the Digital Marketplace.
```
tombye: 'invididually' should be spelt correctly.

tombye: 'invididually' should be spelt correctly here.

PR #64, opened 2015-08-20
Style updates 200815

superroz:


tombye:
Looks like it's just the first commit (to `SQ1-1j-i.yml`) that needs fixing for the tests to pass.

g7_declaration/SQ1-1j-i.yml
```diff
@@ -1,5 +1,6 @@
 question: >
-  If your business was not established in the UK, please confirm if it's a legal requirement in the state where you’re established to provide the services specified in this procurement to be:
+  Please confirm if, to provide the services specified in this procurement, it's a legal requirement in the state where you’re established to be:
+hint: 'You don't need to answer this question if your business was established in the UK.'
```
robyoung: I think the YAML parser is getting confused because this string is quoted but also contains a single quote. Maybe replace the surrounding quotes with double quotes?
```yaml
hint: "You don't need to answer this question if your business was established in the UK."
```

robyoung:
:+1:

PR #65, opened 2015-08-20
Update attachment numbers and hints

superroz:
CCS have updated their attachment numbering. These commits update the numbers and some hint text.


tombye:
:+1:

PR #68, opened 2015-08-21
Update openStandardsSupported.yml

superroz:
added in link to open standards guidance. this will need to be reviewed for style because the latest links in hints aren't showing up properly.


tombye:
I've successfully parsed this using an online markdown editor (http://dillinger.io/) so it gets my :+1:

PR #69, opened 2015-08-25
Correct link URL to be local, not absolute

tombye:
This was missed by the review of the [original pull request](https://github.com/alphagov/digitalmarketplace-ssp-content/pull/67).


quis:
:+1:

PR #70, opened 2015-08-26
Add email validation error messages

TheDoubleK:
I wasn't sure about "You must enter..." vs. "Please enter...", but all the other declaration question messages (e.g. VAT number, DUNS number) use "You must..." so I went with that for emails too.


robyoung:
:+1: I've just noticed that we don't have any versioning on this repo.

PR #72, opened 2015-08-27
Mark optional questions as optional

TheDoubleK:
These are optional questions but were not marked as such, which will prevent users from being able to mark their service as complete unless they fill-in these optional questions.


robyoung:
:+1:

PR #73, opened 2015-08-28
[#102170518] Update supplier pack links

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/102170518

The location of the supplier pack has changed - this updates links to the new location.


quis:
:+1:

PR #74, opened 2015-08-28
Validation message updates

TheDoubleK:
Following discussion with @superroz:

* Change the passive "This question requires an answer" to the active "You need to answer this question."
 * Add new "You need to confirm how you can back up your response." message for security assurances.


ralph-hawkins:
:+1:

PR #77, opened 2015-09-09
Bit of a tidy up

TheDoubleK:
This removes some unused stuff from content:

* VAT number validation message, as we no longer validate VAT numbers;
* `assuranceApproach: null` from Data Storage questions.


quis:
:+1:

PR #79, opened 2015-09-10
Restructure content ready for multiple frameworks

quis:
This commits proposes a new structure for this repo, based on the upcoming requirement to support multiple frameworks.

Part of https://www.pivotaltracker.com/story/show/103088236

```
┣ ./
┃ ┣ tests/
┃ ┗ frameworks/           <- content is separated from tests, README, etc
┃   ┗ g-cloud-7/           <- frameworks are not differentiated by version
┃     ┣ questions/
┃       ┣ services/        <- questions are still split between services/declaration
┃       ┗ declaration/
```

Forthcoming additions to this structure would probably look like this:

```
┣ ./
┃ ┣ tests/
┃ ┗ frameworks/
┃   ┣ g-cloud-7/
┃     ┣ questions/
┃       ┣ services/
┃       ┗ declaration/
┃     ┣ manifests/
┃     ┣ schemas/
┃     ┗ search_mappings/
┃   ┗ digital_services-3/
┃     ┗ …
```

This pull request also changes the tests to check any YAML files in the frameworks directory, agnostic of framework or type of question.


TheDoubleK:
It is possible (probable?) that the questions will change at least slightly between iterations of frameworks.  Are we just going to rely on Git SHAs for versioning? I know that's what we've been doing up until now... I don't know if there's value in freezing a version of the questions for a version somewhere other than in the history?  e.g. say a couple of questions are removed going from G7 to G8 - the latest content will no longer have the field names for fully displaying the G7 services.

TheDoubleK:
I can see the appeal of pushing the schemas out here in future, but I'm not convinced it's the right place for them - they feel like something the API should own.

quis:
@TheDoubleK Versioning is something we should do in the future I think. Same way we do it for utils/toolkit.

tests/test_all.py
```diff
@@ -24,20 +24,16 @@ def test_questions_match_schema(all_files):
...
-            filepath = os.path.join(g6_root_dir, filename)
-            all_paths.append(filepath)
-    for filename in os.listdir(g7d_root_dir):
-        if filename.endswith('.yml') and filename != 'index.yml':
-            filepath = os.path.join(g7d_root_dir, filename)
-            all_paths.append(filepath)
-    return all_paths
+def get_all_files():
+
+    def all_files():
```
robyoung: This seems wrong. Why do we return a function here?

quis: So the returned function can be used as a generator. First time I've done it though, there could be a better way?

robyoung:
It would be good if we could get a convention around the framework directory name so that it matches up with the slug in the frameworks table.

quis:
@robyoung What does the slug look like? `g-cloud-7`?

tests/test_all.py
robyoung: Does it need to be a generator? If it does you could call the function in the outer function to return the generator rather than the function.

quis: @robyoung Changed in f9bc244.

quis:
The frameworks table currently:

`framework` | `framework_slug`
---------------|---------------
`gcloud` | `g-cloud-6`
`gcloud` | `g-cloud-7`

Options we have for making this and the content consistent:

## 1. Content is always the latest, old frameworks get updated with new content

**Folder structure (unchanged)**
`/frameworks/g-cloud/` → `/frameworks/g-cloud/`

**Frameworks table**
`framework: gcloud` → `framework: g-cloud`

## 2. Content is tied to framework iteration, old frameworks can be updated independently

**Folder structure**
`/frameworks/g-cloud/` → `/frameworks/g-cloud-6/`, `/frameworks/g-cloud-7/`, `…`

**Frameworks table (unchanged**
`framework_slug: g-cloud-7` → `framework_slug: g-cloud-7`

## 3. As per 2., but modelling the iterations separately

**Folder structure**
`/frameworks/g-cloud/` → `/frameworks/g-cloud/6/`, `/frameworks/g-cloud/7/`, `…`

**Frameworks table**
`framework: gcloud`, `slug: g-cloud-7` → `framework: g-cloud`, `iteration: 7`

## 4. Other ideas???

minglis:
2 things to add to discussion (not answers but just to muddy waters)

1) If framework iterations at some point don't overlap then we never have too concurrent iterations on the site.

2) If we had to discover what question was asked at some point in time to see the intent of the answer - for example we had an answer of "true" to field "dataExtractionRemoval" and we want to get the full question to discover what we meant by that, do we need to be able to tie it together? This might not need to be in a tool - but we may need to do it.

quis:
_I think this is ready to merge now._

_Changes based on this afternoon's discussion:_

## Name frameworks by iteration
This is so that we can develop new framework iterations in isolation of existing, live frameworks.

## Move questions into their own directory
This is in anticipation of having additional things in the content repository, eg manifests

PR #80, opened 2015-09-21
Add manifests for G-Cloud 6 and 7

quis:
## Add manifest files

The commit extracts the manifest files from the apps into one central place without modifying them.

Although each file is typically only used by one app (at the moment) this will make them easier to keep organised if/when we have 2+ frameworks.

## ~~Add G-Cloud 6 framework as symlink~~

~~This commit adds the G-Cloud 6 framework as a symlink.~~

~~It could be real files in the future, but since we want to keep parity between 6 and 7 it seems reasonable not to maintain two separate collections of files.~~

~~This allows us to access the content for a G-Cloud 6 service page in a dynamic way.~~

## Add G-Cloud 6 as copy of G-Cloud 7

Excludes things which we don't have or need content for, ie declaration, submissions

### DIFF
--- a/frameworks/g-cloud-6/manifests/display_service.yml
+++ b/frameworks/g-cloud-6/manifests/display_service.yml
@@ -1,0 +1,177 @@
+-
+  name: Support
+  questions:
+    - supportTypes
+    - supportForThirdParties
+    - supportAvailability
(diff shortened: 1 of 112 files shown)


allait:
:+1:

PR #81, opened 2015-09-24
DOS 1.0

quis:
## Add minimum required files
These are the bare minimum files that we need so that a user can:
- make a service
- complete a declaration

## Rename DS3 to "Digital outcomes and specialists"

It's official now?

Also, every framework in the frameworks table has an iteration, which I presume should be 1-indexed. Therefore the folder also needs to be named with the iteration to match.

--

Relates to https://www.pivotaltracker.com/story/show/103088236


TheDoubleK:
`digital-outcomes-and-specialists-1` is quite a mouthful...

But :+1: all the same

PR #82, opened 2015-09-28
Add versioning to content repo

quis:
This commit:
- versions the content at 0.0.1, storing the version number in `./VERSION.txt`
- copies the script from the frontend toolkit that tags the repo, based on the contents of the `VERSION.txt` (intended to be run from Jenkins or similar)

image


robyoung:
:+1:

PR #83, opened 2015-10-09
Rename DOS framework

robyoung:
Rename `digital-outcomes-and-specialists-1` to `digital-outcomes-and-specialists`. As Kev says, you only need the number on the sequel.


TheDoubleK:
:+1: - better check with @pcraig3 if he's started using this with the `1` yet?

quis:
Ironically, this is a breaking change and should mean the version is bumped to `1.0.0`

TheDoubleK:
It's only breaking if something is using it already?

PR #90, opened 2015-10-20
Test ALL questions and manifests against schemas

quis:
## Before (not checking all the files)
image

## After
image

This commit
- adds a schema for manifest files
- rewrites the tests so that a schema for messages could easily be added in the future

This also changes the tests to use `pytest.parameterize` instead of a loop inside one test. This means that the tests report how many files they are checking, where before they could look to be passing but actually not reading any files.


robyoung:
:sparkling_heart:  :+1: :sparkling_heart:

PR #91, opened 2015-10-21
Remind users they can edit their declaration later

quis:
image

--

This is a requirement of [the story](https://www.pivotaltracker.com/story/show/105181036).

@superroz can you check the wording?


superroz:
Looks good to me. Might be worth adding the actual deadline info when we know what it is.

PR #98, opened 2015-10-21
Fix spelling mistake

quis:
Ran all the content through a spell checker. This is the only legitimate one it found.


TheDoubleK:
Sterling work :+1:

PR #99, opened 2015-10-21
Lowercase lot values

allait:
Related to https://github.com/alphagov/digitalmarketplace-api/pull/273 and should be deployed to all apps before the API can be merged

### Lowercase lot values
Lot slugs are now all lowercase when returned and accepted by the
API.

### Add uppercase and lowercase lot slugs to question depends
Supporting both lowercase and mixed-case lot slugs makes sure the
apps are working during the API lot changes deployment.

Only lowercase slugs should be used from now on and the mixed-case
abbreviations could be removed once the apps have transitioned.

### Add lotName question
Since the new lot slugs are lowercased they shouldn't be displayed
by the frontend apps. Instead, we use the new service.lotName field
for display and create a mock question to map to that field.

### DIFF
--- a/frameworks/g-cloud-6/questions/services/lotName.yml
+++ b/frameworks/g-cloud-6/questions/services/lotName.yml
@@ -1,0 +1,2 @@
+question: 'Service type'
+type: text
--- a/frameworks/g-cloud-6/manifests/edit_service.yml
+++ b/frameworks/g-cloud-6/manifests/edit_service.yml
@@ -3,7 +3,7 @@
   editable: false
   questions:
     - id
-    - lot
+    - lotName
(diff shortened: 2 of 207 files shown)


frameworks/g-cloud-6/questions/services/lotName.yml
```diff
@@ -0,0 +1,2 @@
+question: 'Service type'
```
TheDoubleK: This is very similar to the `serviceTypes` question - why not something like `Lot name` for these dummy questions?

TheDoubleK:
I've closely reviewed all 207 changed files.
:+1:

PR #102, opened 2015-10-22
Move declaration questions into services

quis:
## Move questions that didn't belong in leftovers

These questions were mistakenly in the 'leftovers' section. Based on the original document this moves them to the places where they should live in the declaration proper.

## Move lot-specific questions out of declaration 

- move these questions in the filesystem and from one manifest to the other
- add the "depends" key to indicate which lots they should be shown for

## Delete the 'leftovers' section

:tada:


frameworks/digital-outcomes-and-specialists/questions/services/bespokeSystemInformation.yml
```diff
@@ -1,3 +1,8 @@
 question: >
-  If you build bespoke systems, do you recognise, and can you ensure, that the data will continue to be available for future government use. 
+  If you build bespoke systems, do you recognise, and can you ensure, that the data will continue to be available for future government use.
 type: boolean
+depends:
+  - on: "lot"
```
allait: `"on"` is the part we need to quote because of yaml's boolean conversion.

(This was ignored by tests because of an error in questions schema, I'm adding a fix in a new commit)

tombye:
:+1:

quis:
Sorry @allait, accidentally rebased out your commit and couldn't work out how to merge it back in. I've added the schema fixes you did into e01b52b, which fixes the `on`s.

allait:
:+1:

PR #106, opened 2015-10-26
Update declaration.yml

superroz:
Sarah C asked to moved the question.


quis:
@superroz I've already moved that question in https://github.com/alphagov/digitalmarketplace-frameworks/pull/102/files but it isn't merged yet. I'll update it to move that question into the same place as in this pull request.

superroz:
Oh super! Thanks.

PR #108, opened 2015-10-28
Add digital-outcomes-and-specialists lot descriptions

allait:
Initial descriptions copied from the digitalmarketplace-prototype.
Lot question is used to render the lot selection page with lot
descriptions for each lot.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists/questions/services/lot.yml
+++ b/frameworks/digital-outcomes-and-specialists/questions/services/lot.yml
@@ -4,12 +4,32 @@ options:
   -
     label: Digital outcomes
     value: digital-outcomes
+    description: >
+      Provide teams to build and support digital outcomes.
+
+
+      Examples of digital outcomes include a discovery phase or an online billing application.
   -
     label: Digital specialists
     value: digital-specialists
-  - 
+    description: >
+      Provide individual specialists to work on digital services, programmes or projects.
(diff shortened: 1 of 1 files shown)


frameworks/digital-outcomes-and-specialists/questions/services/lot.yml
```diff
@@ -4,12 +4,24 @@ options:
...
-  - 
+    description: >
+      Provide specialists to work on a digital service, programme or project. This work must have
+      a defined scope and deliverables. Examples of digital specialists include content designers or
+      technical architects.
+  -
     label: User research studios
     value: user-research-studios
+    description: >
+      Add studios for user research sessions and workshops.
```
quis: This is two paragraphs in the prototype. Might be worth asking @ralph-hawkins how much we care about this because it’ll be extra work to markdownify the template.

quis:
These have changed now: http://dm-prototype.herokuapp.com/states/services-dos

I would suggest we get the latest content in this PR and worry about the Markdown thing later down the line.

allait:
@quis updated the descriptions. I agree that we could add the Markdown thing later if it becomes important.

quis:
Two line breaks for a paragraph in Markdown :sorry:

quis:
:+1: when the tests pass

PR #110, opened 2015-11-02
Add serviceName question for user research studios

allait:
Sets serviceName to be the first question so that every created
draft has a name to display in the service list, similar to G7.

Question is copied from the current prototype.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists/questions/services/serviceName.yml
+++ b/frameworks/digital-outcomes-and-specialists/questions/services/serviceName.yml
@@ -1,0 +1,17 @@
+question: What is the name of the studio?
+type: text
+hint: A studio is the room in which the participant(s) and moderator(s) carry out testing. The name will help buyers refer to the right studio when communicating with you.
+
+depends:
+  -
+    "on": lot
+    being:
+      - user-research-studios
+
+validations:
+  -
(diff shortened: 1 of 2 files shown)


frameworks/digital-outcomes-and-specialists/questions/services/serviceName.yml
```diff
@@ -0,0 +1,17 @@
+question: What is the name of studio?
```
TheDoubleK: Should it be "What is the name of **the** studio"?

frameworks/digital-outcomes-and-specialists/questions/services/serviceName.yml
```diff
@@ -0,0 +1,17 @@
...
+type: text
+hint: A studio is the room in which the participant(s) and moderator(s) carry out testing. The name will help people buyers refer to the right studio when communicating with you.
```
TheDoubleK: This should be "people" or "buyers", but not "people buyers".

Unless we're doing a framework for human trafficking? :trollface:

allait:
@TheDoubleK good call, I've copied the text from the prototype and haven't checked it. Updated.

quis:
:+1:

PR #111, opened 2015-11-04
Employment questions

superroz:
Added 2 new questions from JW and deleted the employment question.


quis:
I think he prefers J-Dubz to JW, but :+1:

PR #113, opened 2015-11-09
Cleanup question keys

allait:
### Remove filterLabel where it maches the question or option label
Removing all instances of filterLabel where it matches the question
name or the option label. This leaves no instances of top-level
filterLabel keys, so they're removed from the question schema for now.

The only question where option labels are different from filterLabels
is `dataCentreTier`. These remaining `filterLabel` keys are renamed to
`filter_label` to make question keys distinct from service data keys.

`filterLabel` keys are currently not used by the frontend apps.

### Rename listItemName to list_item_name
`listItemName` is not currently used by the frontend apps, so the key
can be safely renamed to make it different from the service data keys.

### Remove unused addthing question key
`list_item_name` is used by the list-entry template and js plugin
instead of the `addthing`.

### Fix 2answers-type2 assurance approach typo
2answers-type2 assurance options have 3 possible answers that match
3answers-type2. Replacing all 2answers-type2 values with 3answers-type2
means the 2answers-type2 assurance approach can be removed from supplier
frontend.

### Rename maxLengthInWords to max_length_in_words
Renames the question key to keep the naming consistent. This requires
a small change to the supplier and admin apps, otherwise the word
counter in description fields will disappear.

#### Remove non-existent assuranceHint key from question schema


pcraig3:
Once tests pass.

:+1:

quis:
:-1: Until it includes a version bump.

PR #114, opened 2015-11-10
Add generate-schemas script to create JSON schemas from content files

allait:
Script loads the question files for the given lot and generates a schema
using predefined property templates for different question types.

Current limitations and workarounds:

* Script is using dmutils ContentLoader to get a list of questions and
  `edit_submission` manifest is assumed to contain the full list of
  service data questions
* Pricing fields are treated differently when added to the list of
  required keys
* Assertion answer sets are hardcoded and need to be updated if the
  content changes
* Question max length and word length limits are parsed from the
  question validations. This should be replaced with `max_length`
  and `max_length_in_words` question properties eventually.

Script outputs a JSON file with sorted keys and "required" list.
Script output matches the current G7 schema files, apart from the
added "minItems" and "minLength" records that were skipped in the
G7 schemas.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+docopt==0.4.0
+git+https://github.com/alphagov/digitalmarketplace-utils.git@13.2.0#egg=digitalmarketplace-utils==13.2.0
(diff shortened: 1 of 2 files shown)


robyoung:
There's quite a lot going on here, it would be good to have some tests.

robyoung:
After chatting in real life we decided this is a hard thing to test effectively while the different parts are in different places. I have added a spike into the icebox for reorganising where validation and content lives.
:+1:

PR #116, opened 2015-11-13
Update supplier dec content

sarahonetui:
Updates to content based on feedback from CCS.


quis:
Looks good to me.

quis:
Can you [edit VERSION.txt here](https://github.com/sarahonetui/digitalmarketplace-frameworks/blob/update-supplier-dec-content/VERSION.txt) to be `0.3.1`?

TheDoubleK:
Are the questions `subcontractorInformation1` and `subcontractorInformation2` still specific to subcontractors?  It looks like they have been made more generally applicable, in which case the names of the files (and so the keys for the saved information) should probably change to not mention `subcontractor`.

sarahonetui:
Good point Kev, have renamed them. Thanks

quis:
:+1:

PR #118, opened 2015-11-17
Fix example of outcomes

quis:
This was changed for the `open` state; it should also have been changed for the `coming` state.


TheDoubleK:
:+1:

PR #119, opened 2015-11-18
Add specialist questions

allait:
#### Update schemas with new question and manifest keys
Adds multiquestion question type and `name`, `questions`, and
`empty_message` fields.

Adds additional schema validation to check that `questions` is only
valid for `multiquestion` questions, and `fields` is only valid for
`pricing`.

Adds `edit_questions` key to the manifest schema.
7647a68
 @allait	allait	Add section keys to the README


quis:
Pretty clean :+1:

scripts/generate_dos_specialists_questions.sh
```diff
@@ -52,3 +68,6 @@ make_questions "service manager" "serviceManager"
 make_questions "technical architect" "technicalArchitect"
```
pcraig3: Do we care if they're in alphabetical order?

allait: Doesn't matter for the script since it only creates the question files, so order would only help with readability. Manifest order matters since it's the order of service questions, so I kept the order from the spreadsheet. Might be worth asking Ralph or Sarah if we want alphabetical in the manifest

pcraig3: Oh, right.  Duh >.<

pcraig3:
Just asked @sarah-cunliffe and she says we want alphabetical specialists and capabilities.

allait:
@pcraig3 sorted specialists

PR #120, opened 2015-11-19
Revert DOS address field in declaration

robyoung:
[#107994592] https://www.pivotaltracker.com/story/show/107994592

We want to make sure we get relatively clean address data.


quis:
:+1:

PR #121, opened 2015-11-19
[#108525606] Add Quality assurance analyst to specialist roles.

pcraig3:
Pretty straightforward.  Added a line to the `generate_dos_specialists_questions.sh` file and generated everything magically.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/108525606)

Also added some stuff to our `requirements.txt` files so that they install properly.
Not really related, but didn't seem an important enough change to do separately.


robyoung:
:+1:

PR #122, opened 2015-11-20
[#107188868] Split lab address into three separate fields

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/107188868

The story says to split the address into three fields, but the prototype page linked to has it over four fields (https://dm-prototype.herokuapp.com/states/studio-location)

I've gone with three, but we can iterate to four if necessary:
screen shot 2015-11-20 at 10 28 41


quis:
The guidance on the service manual says four fields: https://www.gov.uk/service-manual/user-centred-design/resources/patterns/addresses.html#multiple-fields

However since we can’t do the thing where the first two fields are grouped with one label, I think that this is the best solution.

:+1:

PR #123, opened 2015-11-20
Some validations for dos questions

TheDoubleK:
This adds some validations to the DOS service questions:
 * Length limits of 50 words/500 characters on large textboxes
 * Length limits of 20 words/200 characters for address lines
 * Length limit of 10 characters for a postcode
 * Length limit of 100 characters on how many people a lab can accommodate
 * `answer_required` on boolean questions that didn't previously have it

I've also added some hint text taken from the prototype on a couple of questions.

I haven't done anything with:
 * multi-questions (I think validations for these should be at the sub-question level?)
 * prices (I know @tommorris has been working on these)
 * locations (they don't feel like they are required - e.g. for remote working only)

### DIFF
--- a/frameworks/digital-outcomes-and-specialists/questions/services/anonymousRecruitment.yml
+++ b/frameworks/digital-outcomes-and-specialists/questions/services/anonymousRecruitment.yml
@@ -4,3 +4,7 @@ depends:
   - "on": "lot"
     being:
       - user-research-participants
+
+validations:
+  - name: answer_required
+    message: 'You need to answer this question.'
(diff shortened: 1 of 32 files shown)


pcraig3:
I know it's probably more widespread than this pull request (and probably not a very important problem), but is there a reason we're wording our error messages differently?

ie: 
- Your answer must be no more than ...
- Your answer can't be more than ...

TheDoubleK:
Because of cutting-and-pasting from different places...?

TheDoubleK:
Done - I've gone with `must be no more than` because it's in the majority and I'm sure that was the one that @superroz settled on.

I've also fixed G6 and G7 messaging to all be the same, so next time someone copies and pastes a validation message it will be in the same style.

pcraig3:
Seems okay to me :+1:

PR #124, opened 2015-11-20
Update supplier dec content

sarahonetui:
Removing sufficientInformation as covered by other questions
Updated firstRegisteredCountry -> currentRegisteredCountry based on CCS feedback


robyoung:
:+1:

PR #127, opened 2015-11-23
[#107188910] Generate JSON schemas which validate "at least one of" a group of questions.

pcraig3:
Generates JSON files to correctly validate services where a "at least one of" a group of questions have to be present.  For example, at least one specialist needs to be present for a service to validate, and a specialist consists of more than one related keys.

Needs pull request [#213](https://github.com/alphagov/digitalmarketplace-utils/pull/213) in utils to pass tests.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/107188910)

##### Structure of the JSON schema

As before, all of the possible keys a schema will accept need to be defined in the top-level `"properties"` dictionary, and then groups of related keys must be specified in their own schemas in under a top-level [`"anyOf"` key](http://json-schema.org/latest/json-schema-validation.html#anchor85). 

Kind of looks like this:

```json
{
  "properties": { 
    "thing": {
      "type": "string"
    },
    "subthing_1": {
      "type": "string"
    },
    "subthing_2": {
      "type": "string"
    }
  },
  "anyOf": [
    {
      "required": [ "subthing_1", "subthing_2"],
      "title": "specialist"
    }
  ]
}
```

Also added a (not strictly necessary) `"title"` key to the `"anyOf"` subschemas so that we can return it when an `"anyOf"` validation fails.  The default error message that comes out of the JSON schema validator isn't really that helpful.

### DIFF
--- a/scripts/generate-schemas.py
+++ b/scripts/generate-schemas.py
@@ -5,7 +5,6 @@ Usage:
     generate-schemas.py --output-path=<output_path>
 
 """
-from collections import defaultdict
 
 import os
 import re
@@ -111,7 +110,14 @@ def list_property(question):
     }}
 
 
+def pricing_fields_property(question):
+    prop = text_property(question)
+    return {val: prop[question.id] for val in question.fields.values()}
+
+
 def pricing_property(question):
+    if question.get('fields'):
+        return pricing_fields_property(question)
     return {
         "priceMin": {
             "type": "string",
@@ -179,7 +185,7 @@ QUESTION_TYPES = {
     'radios': radios_property,
     'boolean': boolean_property,
     'list': list_property,
-    'pricing': text_property,
+    'pricing': pricing_property,
     'percentage': percentage_property,
     'multiquestion': multiquestion
 }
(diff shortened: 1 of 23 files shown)


scripts/generate-schemas.py
```diff
@@ -50,6 +50,7 @@ def empty_schema(schema_name):
         "type": "object",
         "additionalProperties": False,
         "properties": {},
+        "anyOf": [],
```
allait: We probably don't need to add the key here since we're just going to remove it if it's empty

scripts/generate-schemas.py
```diff
@@ -267,14 +284,40 @@ def build_question_properties(question):
...
 
+def build_any_of(any_of, fields):
+    return {
+        'required': [field for field in sorted(fields)],
+        'title': any_of
+    }
+
+
 def build_schema_properties(schema, questions):
+    any_ofs = {}
```
allait: Why is `any_ofs` a dictionary? It seems that storing subschemas indexed by question_id with `any_ofs[quesiton.id]` doesn't really add anything, so we might as well start with the list and skip the dictionary sorting and conversion to list

allait:
:+1:

scripts/generate-schemas.py
pcraig3: Yup, fair point.

scripts/generate-schemas.py
pcraig3: Dictionary converts quite easily to a sorted list is all.  Otherwise, the order isn't guaranteed.

PR #128, opened 2015-11-24
Make capability and specialists optional questions with schema dependencies

allait:
### Make specialist questions optional
At least one specialist section is required, but each individual
question can be left empty.

Multiquestions are marked as optional to make sure frontend doesn't
display the "Answer required" field and schema generation doesn't
add the question fields to the required list, but individual location
and pricing questions don't have optional set to true, so property
schemas for checkboxes contain the correct `minItems: 1` constraint.

### Add multiquestions for team capabilities
Each capability should be edited on a separate page, so capability
questions are wrapped with a multiquestion that can be rendered as
section.

### Add answer_required messages to individual specialist questions
Even though the questions are optional, API should return an error
if only one of the questions was answered when the page was edited.

### Add multiquestion dependencies to schema generation
To validate that all questions for a multiquestions are answered
at the same time we use the JSON schema "dependencies" property
and for each possible question field set the remaining ones as
being required if the question was answered.

### Fix pricing field schema generation
Each pricing field has specific schema requirements, so the overall
template cannot be replaced by text fields


frameworks/digital-outcomes-and-specialists/questions/services/businessAnalystDayRate.yml
```diff
@@ -7,3 +7,7 @@ type: pricing
 fields:
   minimum_price: businessAnalystPriceMin
   maximum_price: businessAnalystPriceMax
+
+validations:
+  - name: answer_required
```
robyoung: These price ones won't work but that's fixed in https://github.com/alphagov/digitalmarketplace-frameworks/pull/130

robyoung:
Looks like it should work, generates something that looks reasonable locally :+1:

pcraig3:
For our digital-specialist multiquestions, we've included an [`any_of`](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/questions/services/agileCoach.yml#L8) key so that the API can return an error message that's meaningful.  Probably we should use one for digital-outcomes as well.

TheDoubleK:
:+1:

PR #126, opened 2015-11-20
Adding some fields for pricing

tommorris:
This change makes use of the updates in https://github.com/alphagov/digitalmarketplace-utils/pull/210 to implement pricing questions with `fields`.

This also updates the schema generation to support the new style of pricing definition.

It is a major version bump because it changes the structure of pricing questions and requires dmutils and the frontend toolkit to also be upgraded.

See:
- https://github.com/alphagov/digitalmarketplace-utils/pull/210
- https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/188

PR #130, opened 2015-11-25
Fields for pricing on dependencies

robyoung:
Rebase of https://github.com/alphagov/digitalmarketplace-frameworks/pull/126 on top of https://github.com/alphagov/digitalmarketplace-frameworks/pull/128

This change makes use of the updates in https://github.com/alphagov/digitalmarketplace-utils/pull/210 to implement pricing questions with `fields`.

This also updates the schema generation to support the new style of pricing definition.

It is a major version bump because it changes the structure of pricing questions and requires dmutils and the frontend toolkit to also be upgraded.

See:
- https://github.com/alphagov/digitalmarketplace-utils/pull/210
- https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/188


TheDoubleK:
All looks good - makes sense :+1:

PR #137, opened 2015-11-25
service essentials: added short names and combined a question

ralph-hawkins:
* Service essential questions now have short versions, that appear on the service summary page and in the error messages.
* Removed question that was specifically about publishing software as this was confusing for users that don't directly work on software (eg user researchers).
* Combined the important bit of the removed question with the question open standards principles


allait:
:+1:

PR #138, opened 2015-11-25
[#108938292] Update DOS closing date

robyoung:
https://www.pivotaltracker.com/story/show/108938292


pcraig3:
`You can come back and change your answers at any time before 19 January 2016.`

Do we need to explain that they can also change them up until 3 p.m., or is that close enough?

quis:
I think we should be consistent and mention the time everywhere.

quis:
Someone might argue that you also need the time zone. That person isn’t me. So :+1: if you are at peace with your conscience.

PR #141, opened 2015-11-26
Remove extra type from recruit method

ralph-hawkins:


VERSION.txt
```diff
@@ -1 +1,5 @@
-0.9.2
\ No newline at end of file
+<<<<<<< HEAD
```
allait: :fearful:

allait:
:+1:

PR #142, opened 2015-11-26
Add default values to labPrice 

pcraig3:
Not having a default unit cause the [`question.value`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/content_loader.py#L512) check to fail even when `labPriceMin` and `labPriceMax` were set to permissible values.

***

## Before

screen shot 2015-11-26 at 15 20 24

## After 

screen shot 2015-11-26 at 15 19 20


allait:
:+1:

PR #143, opened 2015-11-27
Making login/create account links on the homepage sidebar relative

tommorris:
Currently the absolute links break on preview - clicking them will take you to production, which means we can't use end-to-end functional testing on those links.


quis:
:+1:

PR #144, opened 2015-11-27
Add dos price validations

robyoung:
- Add DOS price validations to generator
- Add DOS price validations for specialist day rates
- Add price validations for labs


allait:
:+1:

PR #146, opened 2015-11-30
Add qa to manifest

robyoung:


quis:
:+1:

TheDoubleK:
Should this be a major version bump as it will require an update to schemas in the API before it can be used?

quis:
I don’t think so. You could upgrade now without breaking anything. It’s just some extra work to do to _use_ the new feature.

TheDoubleK:
But if you upgrade the content on the frontend then trying to save answers for qualityAssurance will break because they don't exist in the API, won't they?

quis:
That is true. Makes pretty much everything a breaking change though.

TheDoubleK:
Yep.  Pretty much everything in content is :/

pcraig3:
<sub>[oh whoops](https://github.com/alphagov/digitalmarketplace-frameworks/pull/121/files)</sub>

PR #147, opened 2015-11-30
Add optional section descriptions

quis:
To describe or give more context about a section on the summary page.

Screenshot from the prototype:
image


pcraig3:
Seems okay if the distinction between `description` and `summary_page_description` is clear. 
Probably the README is a good place for that.

allait:
New field needs to be added to the README, but other than that :+1:

PR #148, opened 2015-11-30
Lab questions

carolinegreen:


pcraig3:
:+1:

TheDoubleK:
This should probably have been a major version update as it will break things if used by the front-end apps without rebuilding schemas and updating the API.

PR #150, opened 2015-11-30
[#109158740] Add more linebreaks so that markdown filter works properly.

pcraig3:
Basically, the Outcomes locations question wasn't rendering correctly.  

Basically, linebreaks in the question for `outcomesLocations.yml` were being ignored because they weren't broken enough.

So basically, our markdown filter only recognises linebreaks when there's 2 of them consecutively.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/109158740)

***

#### Bad
screen shot 2015-11-30 at 15 59 41

#### Good
screen shot 2015-11-30 at 15 58 36


quis:
For future reference, this web tool matches the syntax of our Markdown filter: http://daringfireball.net/projects/markdown/dingus

quis:
:+1: to making it work for now (both single and double line breaks get parsed into two paragraphs by the Markdown dingus, but our filter only respects the double).

You’ll want to add some spacing between the paragraphs when you bring this change into the supplier frontend.

quis:
https://github.com/alphagov/digitalmarketplace-frameworks/releases/tag/v0.11.2 already exists.

You should make a version number of your very own.

PR #151, opened 2015-11-30
Alter generate schemas script

carolinegreen:
Alter generate schemas script to include hours options.  Joint commit with @carolinegreen and @pcraig3


allait:
:+1:

PR #153, opened 2015-12-01
Limit schema price string to 15 digits before decimal point

allait:
Long price string without spaces break the layout and since there's
no reason to have them we just limit the strings accepted by the
JSON schema to 15 character before the decimal point and 5 after.


pcraig3:
:+1:

PR #154, opened 2015-12-01
Add dates as messages

robyoung:
Ultimately these should be stored in the API but for the moment we don't know how many different dates there will be and it is therefore difficult to know what the right way of managing them is.


allait:
:+1:

PR #156, opened 2015-12-02
Add dummy dates file for g-cloud 6

robyoung:
The dates messages are loaded on the supplier dashboard even though none of them are used for live or expired frameworks.


allait:
:+1:

PR #157, opened 2015-12-03
Add intention to award dates

robyoung:
- These dates are needed on the framework dashboard.
- The timezones have been added as markup and a spike added to the icebox ([#109431100](https://www.pivotaltracker.com/story/show/109431100)) to review how dates are managed


frameworks/g-cloud-7/messages/dates.yml
```diff
@@ -1,4 +1,5 @@
-clarifications_close_date: "5pm, 22 September 2015"
-clarifications_publish_date: "5pm, 29 September 2015"
-framework_close_date: "3pm, 6 October 2015"
-framework_live_date: "23 November 2015"
+clarifications_close_date: '5pm <abbr title="Greenwich Mean Time">GMT</abbr>, 22 September 2015'
```
allait: I think G-Cloud times were in BST

tombye:
Only point is I'd say this should be `0.13.0`, not `0.12.2` as it's a feature you're adding here, not a bug fix.

schemas/messages.json
```diff
@@ -12,6 +12,7 @@
         "clarifications_close_date": {"type": "string"},
         "clarifications_publish_date": {"type": "string"},
         "framework_close_date": {"type": "string"},
+        "intention_to_award_date": {"type": "string"},
```
allait: Is the new field optional or should it be added to the required list?

frameworks/g-cloud-7/messages/dates.yml
tombye: Talking to @superroz, we're out of BST and back into GMT now so this is actually correct.

For reference: http://www.timeanddate.com/time/united-kingdom-bst.html

Note: I wish we were still in BST :(

allait: But that time isn't right then. We've closed submission on 3pm, when UK was on BST. So it's either 2pm UTC in October, or 3pm BST, since that's what the text said when G-7 was open.

PR #158, opened 2015-12-03
Content updates

ralph-hawkins:
* added extra intro to essentials
* renamed Location to Transport on the labs summary
* moved the locations you can recruit to become the last question for user research participant recruitment
* made the do you provide wifi a yes/no
* changed the order of the questions where options where yes as standard/yes for additional cost/no - so No is last
* Added a link to a map for all the questions with locations
* Other minor content updates


robyoung:
I'm looking at getting the tests passing here.

frameworks/digital-outcomes-and-specialists/manifests/edit_submission.yml
```diff
@@ -125,3 +126,11 @@
   questions:
     - recruitMethods
     - recruitFromList
+
+- name: Location
+  editable: True
+  description: >
+    Help buyers understand how easy it is for visitors to reach your studio.
```
allait: Should this be the same as the "Transport" description?

allait:
:+1:

PR #159, opened 2015-12-03
Add "Apply to provide" to dos lot links text

allait:
Changes the lot question option labels to match the prototype text.
This means that lot question labels no longer match lot names, but
since the quesiton is only used for the submission lots page this
doesn't affect any other pages at the moment.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists/questions/services/lot.yml
+++ b/frameworks/digital-outcomes-and-specialists/questions/services/lot.yml
@@ -2,7 +2,7 @@ question: 'Service type'
 type: radios
 options:
   -
-    label: Digital outcomes
+    label: Apply to provide teams for digital outcomes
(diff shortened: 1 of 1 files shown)


TheDoubleK:
I have a feeling these are the `label`s used for the "are you sure you want to delete..." messaging when deleting a lot.

e.g. here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/services/_base_service_page.html#L38

allait:
@TheDoubleK I think those are for removing outcomes/specialists, so that line is looking up the specialist or outcome name, not the lot. I'm not sure if there's another path to that template, but I don't think `.get_question_by_slug` works on the `lot` question

TheDoubleK:
OK - if you're sure - :+1:

PR #160, opened 2015-12-03
Bump version to 0.12.3

TheDoubleK:
The most recent merge did not include a version number bump, so the tagging job failed.  The latest changes could be pulled in using the commit SHA, but it would be nicer to have a version number for it.

Minor version bump as I can't see any breaking changes or new stuff in the diff here (the last version tag up to most recent merged commit): https://github.com/alphagov/digitalmarketplace-frameworks/compare/v0.12.2...4cf62a686f07042a02123f5d7998ccc0417df8d1.


allait:
:+1:

PR #161, opened 2015-12-03
Descriptions for capabilities and specialists

ralph-hawkins:
* added descriptions for all the specialists roles
* added descriptions for all the capability types


robyoung:
I think that adding these hints and descriptions makes the question generation script no longer useful.

TheDoubleK:
The hints and descriptions seem like they will be useful in the app... should we get them in and then try to fix the script?  It might be OK if we don't have the CSV right away - we can post it as a communication early next week if we can't fix the script today?

frameworks/digital-outcomes-and-specialists/questions/services/designer.yml
```diff
@@ -12,3 +12,5 @@ questions:
   - designerDayRate
 
 empty_message: You haven't added a designer
+
+hint: Set up a team for successful delivery. Remove obstacles, track progress, facilitate meetings and help the team organise itself.
```
quis: This one is definitely wrong.

frameworks/digital-outcomes-and-specialists/questions/services/businessAnalyst.yml
```diff
@@ -12,3 +12,5 @@ questions:
   - businessAnalystDayRate
 
 empty_message: You haven't added a business analyst
+
+hint: Specify, collect and present key performance data and analysis for their service.
```
quis: Finding out if this is the right description…

quis: @sarahonetui is looking into it…

frameworks/digital-outcomes-and-specialists/questions/services/designer.yml
quis: In the doc it is
> Design interactions that support user needs. This may involve sketching, design production, hand-coded HTML and CSS, and rapid prototyping.

…but this doesn’t cover service design and graphic design. We’re going to come up with an alternative.

robyoung:
@TheDoubleK I wasn't suggesting not putting this in. It was more that we should disable or delete the script as it will clobber all these questions if it's run now. It's not a big deal though, people can just not use the script until we fix it.

pcraig3:
I've talked about it with @sarahcunliffe and we've agreed how to spit out the questions.
Good to merge once the copy is agreed upon. :+1:

PR #163, opened 2015-12-04
Don’t hyphenate off-site

quis:
> Roz Strachan [14:11]
> Rats. Why is off-site hyphenated?
>
> ​[14:11]
> That's really weird.
>
> ​[14:12]
> Offsite and onsite should both be one word.
>
> ​[14:12]
> Is that something we can change?
>
> Chris Hill-Scott [14:13]
> It might make Kev angry…
>
> Roz Strachan [14:13]
> I'll take that hit. We never have enough time.


TheDoubleK:
At least it's not all lowercase :+1:

PR #165, opened 2015-12-04
Add linebreaks for markdown interpreter

pcraig3:
Our `markdown` filter doesn't insert newlines unless it sees two consecutive newlines. 

So I've added several where @sarahcunliffe pointed them out.


quis:
:+1:

PR #167, opened 2015-12-04
Add section descriptions for service declarations

quis:
We need to be clear about your answers to these questions.

Something like this was on the service page in the prototype, but this:
- makes more sense
- is more official


robyoung:
:+1:

PR #168, opened 2015-12-04
Changed locations The Midlands to West Midlands and East England to E…

ralph-hawkins:
…ast of England, to match the ONS map we link to


robyoung:
:+1:

PR #180, opened 2015-12-08
Add correct Service desk hint text

suedavis68:
Added hint text for Service desk, original hint text duplicated Network administration hint text


robyoung:
This should probably be signed off by @cathrooney or @sarahonetui

sarahonetui:
@robyoung yes, this is ok thanks. all hint text fixes are fine.

PR #184, opened 2016-01-11
Testing improvements

tommorris:
Unit tests for generate_schema, documentation improvements etc.


scripts/generate_schemas.py
```diff
@@ -2,7 +2,7 @@
 """Generate JSON schemas from the frameworks questions content.
 
 Usage:
-    generate-schemas.py --output-path=<output_path>
+    generate_schemas.py --output-path=<output_path>
```
allait: What's the purpose of the rename? `-` is what we seem to have settled on in `digitalmarketplace-scripts` repo.

tommorris: Because the unit tests need to import functions from the script.

scripts/support.py
```diff
@@ -0,0 +1,29 @@
...
+
+opened_files = []
+
+
+def clear_buffer():
+    global opened_files
+    opened_files = []
+
+
+oldopen = builtins.open
```
allait: Can we use a fixture that doesn't rely on the global variable instead of this module?

tests/test_generate_schema.py
```diff
@@ -0,0 +1,351 @@
...
+    checkbox_property, percentage_property, multiquestion, \
+    build_question_properties, empty_schema, load_questions, \
+    drop_non_schema_questions, radios_property, list_property, price_string, \
+    pricing_property, generate_schema, SCHEMAS
+from scripts import support
+
+
+# Python 3 compatibility
+try:
+    xrange
```
allait: Is there a problem with just using python2 `range`? It looks like we're building a list from it either way

scripts/generate_schemas.py
allait: Let's move the functions out and keep the script as an entry point. It should make it more obvious when we're testing the generation logic as opposed to the script itself.

tests/test_generate_schema.py
tommorris: Fixed.

scripts/support.py
tommorris: I looked into different alternatives and couldn't find any better way to achieve this. This module is only loaded during unit testing, so it won't affect the build process outside of a testing context.

scripts/generate_schemas.py
tommorris: Okay, will refactor this in just a moment.

tommorris: Done.

schema_generator/__init__.py
```diff
@@ -0,0 +1,371 @@
...
+    drop_non_schema_questions(questions)
+    schema = empty_schema(schema_name)
+
+    build_schema_properties(schema, questions)
+    add_multiquestion_anyof(schema, questions)
+    add_multiquestion_dependencies(schema, questions)
+
+    with open(os.path.join(path, 'services-{}-{}.json'.format(framework_slug, lot_slug)), 'w') as f:
+        json.dump(schema, f, sort_keys=True, indent=2, separators=(',', ': '))
+        f.write('\n')
```
robyoung: Brrp, this isn't a unix text file.

robyoung:
Does `scripts` need to be a package? It isn't in other repos.

robyoung:
Doing the moving the existing code to a package and making changes to that code in one commit makes it quite hard to see what's been changed.

schema_generator/__init__.py
```diff
@@ -0,0 +1,371 @@
...
+
+    builder = loader.get_builder(framework_slug, 'edit_submission').filter({'lot': lot_slug})
+    return {q['id']: q for q in sum((s.questions for s in builder.sections), [])}
+
+
+def drop_non_schema_questions(questions):
+    """
+    These questions are used to generate questions in the UI but are not used to validate the response, so we
+    remove them.
+    """
```
robyoung: We try to follow [PEP257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings), at least in so far as function docstrings are commands.

schema_generator/__init__.py
tommorris: I haven't touched it. That's how I found it. ;-)

tommorris:
`scripts` as package: fixed. Will resolve PEP257 shortly.

schema_generator/__init__.py
tommorris: Will switch to a non-docstring comment.

allait:
Made a branch with a commit replacing the `tests/support` and global with a fixture: https://github.com/alphagov/digitalmarketplace-frameworks/compare/testing-improvements...test-improvements-fixtures

It also highlights a problem wit the schema test: 0 length set diff assert passes because the expected file list is empty.

allait:
`.hypothesis/examples.db` shouldn't be commited

scripts/support.py
```diff
@@ -0,0 +1,29 @@
+try:
```
robyoung: This is a test helper, it feels like it should be in the `tests` package.

tests/support.py
```diff
@@ -0,0 +1,29 @@
...
+
+def clear_buffer():
+    global opened_files
+    opened_files = []
+
+
+oldopen = builtins.open
+
+
+def newopen(*args):
```
robyoung: I get a bunch of failing tests until I add keyword arguments to this, specifically `encoding` seems to be used (apparently not in our code though).

tommorris:
Removed `hypothesis.db`. @allait's solution does seem better.

tests/support.py
```diff
@@ -0,0 +1,29 @@
...
+    global opened_files
+    fh = oldopen(*args)
+    opened_files.append(fh.name)
+    return fh
+
+builtins.open = newopen
```
robyoung: It's slightly more involved but this could be achieved without the global variable or the unbound wrapping of a builtin with a couple of closures in a pytest fixture.

```python
@pytest.fixture(autouse=True)
def proxy_open(request):
    old_open = builtins.open
    opened_files = []

    def new_open(*args, **kwargs):
        fh = old_open(*args, **kwargs)
        opened_files.append(fh.name)
        return fh

    builtins.open = new_open

    def undo():
        builtins.open = old_open

    request.addfinalizer(undo)

    return opened_files
```

schema_generator/__init__.py
robyoung: Now seems like a great time to fix it then.

tests/support.py
allait: I think we did the same thing :smile:  https://github.com/alphagov/digitalmarketplace-frameworks/compare/testing-improvements...test-improvements-fixtures

tommorris: Okay, I'll pull @allait's changes into this branch.

allait:
:+1:

PR #185, opened 2016-01-19
Fix missing newline at the end of file PEP8 error

allait:
run_tests scripts in apps don't ignore frameworks files when running
pep8 checks, since frameworks didn't include python files before.

Now that schema_generator is a package it gets picked up by pep8
and fails the check due to missing newline.


TheDoubleK:
:+1:

PR #189, opened 2016-02-10
Add a `Requirements title` question

TheDoubleK:
The first question for building a brief needs to be a title that it can be referred to later.
Words taken from the prototype and agreed with Ralph.

PR #190, opened 2016-02-10
Fix missing EOF newline in generated schemas

allait:
When script was rewritten to use schema_generator the call to add a
trailing newline to generated files was lost, which means rerunning
the script creates schema diff without EOF newline.

Restoring the write to keep the trailing newline in all generated
schemas.


TheDoubleK:
This is also fixed here: https://github.com/alphagov/digitalmarketplace-frameworks/commit/c3a7525e661d7950cd823b3db5f2152eee66ecdb

I propose closing this PR and putting #189 through, as it has other things too.

PR #191, opened 2016-02-10
Fix duplicate brief location options

allait:
Matching location options generate an invalid schema, since we set
`uniqueItems` on the checkboxes property.

Copying the location list from service questions.


pcraig3:
:+1:

TheDoubleK:
This is also fixed in #189...

PR #192, opened 2016-02-11
Use option value in JSON schemas if available

robyoung:
When defining checkboxes or radio sets a value can be provided to each option as well as a label. If provided this is used as the value attribute of the option. This change updates the JSON schemas to use that value rather than the label if the value has been provided.

Props to @allait for showing me how to hypothesis.


pcraig3:
:+1:

PR #193, opened 2016-02-12
Tidy up brief questions

TheDoubleK:
Changes are mainly harmonising manifest `name`s with question `name`s. This is particularly important for single-question pages because any validation messages refer to the question `name` but the page title is the manifest `name`. If these are different then the validation errors refer to different questions from what can be seen on the page.

Also fixes some `Title Case` things to favour just an `Initial capital letter` style.


pcraig3:
Seems :+1: to me.

PR #194, opened 2016-02-12
Hundred word textboxes

TheDoubleK:
The `under_xxx_words` error keys should match the number of words specified by `max_length_in_words` so that they match the errors returned by the API.

These changes do not require a re-generation of the API schemas at all - the API is already returning `under_100_words` validation errors in the right place.  This is just so that the frontend can map these to the correct validation message.


frameworks/digital-outcomes-and-specialists/questions/briefs/backgroundInformation.yml
```diff
@@ -14,5 +14,5 @@ validations:
     name: answer_required
     message: 'You need to answer this question.'
   -
-    name: under_50_words
+    name: under_100_words
     message: 'Your answer must be no more than 50 words.'
```
allait: Message text should match the actual word limit

TheDoubleK: Aargh, sorry - yes I've changed it back and forth a couple of times but missed it the important time :(

allait:
:+1:

PR #196, opened 2016-02-15
[#112338109] Add brief-response questions; introduce `boolean_list` type

pcraig3:
This pull request adds `.yml` question files for suppliers' responses to briefs.  

`essentialRequirements` and `niceToHaveRequirements` are both boolean lists of yes/no questions (entered via text field) buyers need suppliers to answer.  This means that we can't hardcode the questions beforehand because we don't know what they're going to be.

We *do* know that the answers to these questions will be returned as (indexed) lists of booleans, so that's what we're looking for in the JSON schema that the API will use. 

Testing them meant making small additions to the existing test suite.  Other than the new answer type (`boolean_list`), brief-responses are identical to the other question schemas we have, so all of the current tests can easily be reused.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/112338109)


VERSION.txt
```diff
@@ -1 +1 @@
-0.16.4
+0.16.5
```
allait: I think this deserves at least a minor release bump :rocket:

robyoung:
I'm not sure `yes_no_questions` is the best name. We already have a `boolean` type and this is a list of booleans but doesn't give any indication that it's related to the `boolean` type. You only discover that if you look at the schema generator.

pcraig3:
They've changed a bit, actually. 

They (will) now look like:
```python
 ('q5-0', 'true'),
 ('q5-1', 'false'),
 ('q5-4', 'true')
```
and will be returned as a list of booleans and Nones.
So, after processing it, the content loader returns:
```python
{
  'q5': [True, False, None, None, True]
}
```

Point taken though.  If we want them named after what they contain, we could go with something like `boolean_list`.  Thoughts?

robyoung:
Yea, `boolean_list` works.

robyoung:
:+1:

PR #197, opened 2016-02-16
Changes `users` to `endUsers` to avoid key clash

TheDoubleK:
The `users` key clashes with the `users` of a brief, i.e. the list of users who are associated with the brief and allowed to work on it.  This means that answers submitted to this question are overwritten with the user list for the brief in the brief API response.

By making this change to the question key we will avoid the name clash.


pcraig3:
:key: :no_good: :key: 

:+1:

PR #198, opened 2016-02-24
Brief responses should be all in one section

TheDoubleK:
Because all questions have to be responded to on a single page.


robyoung:
:+1:

PR #201, opened 2016-02-26
Add clarification question

robyoung:
Manage clarification question form with the content loader to make API responses and validation more easier and more consistent.


TheDoubleK:
:+1:

PR #205, opened 2016-03-04
Add id to service questions

TheDoubleK:
Question IDs usually come from the filename, but for some cases:
  * G-Cloud serviceTypes
  * DOS locations

 We want to be able to specify the ID to use, different from the filename of the question.


robyoung:
:+1:

PR #207, opened 2016-03-08
Make date format more explicit

pcraig3:
Sue pointed out that when we refer to dates, we should make the days and the months explicit.  It's GDS style or something.

With `01/01/2016` it's not clear.
With `31/12/2016` it is.


carolinegreen:
:+1: Makes sense, people could get confused between British and American dates.

TheDoubleK:
Nice spot.

pcraig3:
@TheDoubleK I wish I could take the credit but it was Sue.

PR #209, opened 2016-03-10
Add file type and size to ZIP file download links

TheDoubleK:
An [earlier commit](https://github.com/alphagov/digitalmarketplace-frameworks/commit/2c747d2133ac1f7fbec34028805f1ea79802f3c4) added file type and size information to PDF download links.

This adds similar link text for the ZIP file downloads that I missed in the earlier commit.

As per @tombye's comment: 
 "If we're linking to a document, it should include the file type and size as on www.gov.uk.

An example of what I mean at the bottom of: https://www.gov.uk/green-deal-energy-saving-measures/improvements-and-benefits-to-your-home"


tombye:
This'll make a big difference to any users on a bad connection :+1:

TheDoubleK:
Well, it would have done when G7 and DOS were open for applications!  Doesn't hurt to have it right though, for future reference.

PR #210, opened 2016-03-21
G Cloud 8

TheDoubleK:
This is a result from the spike to get G-Cloud 8 up and running: https://www.pivotaltracker.com/story/show/115252841

The questions here are identical to G-Cloud 7 questions, both for service submissions and for the declaration. The dates mentioned are purely fictional and will need to be updated with the actual dates once we know them.

We agreed in the spike feedback meeting to add the G-Cloud 8 content to the repository now so that question content can be tweaked before we do the dev work of putting G-Cloud 8 open for submissions.


robyoung:
How do we ensure that we don't forget to change the dates? Can they be obviously fake (not a date or a date in the past)?

TheDoubleK:
All the content will be reviewed before we turn anything on I reckon we'll be fine - people using it on Preview and Staging will pick it up in the weeks beforehand we have scheduled to thoroughly test it all before we have to turn it on :trollface: 

Oh, right you are - I'll do something to make it obvious.

robyoung:
:+1:

PR #212, opened 2016-03-24
Make brief specialist values match the service specialist keys

allait:
DOS services store specialist location and day rate using a camelCase version of the role name: eg agileCoachLocations. Since the brief specialist value is used to build the location key to use for filtering in the API, the role value should match the service key prefix to avoid any rewriting logic in the API.


robyoung:
:+1:

PR #213, opened 2016-03-24
Brief responses for csv

TheDoubleK:
This adds a manifest that is used to determine columns in the csv file of brief responses.

It also adds `respondToEmailAddress` to brief responses in all lots, as we need to tell the buyer how they should get in touch with the supplier about their application.

It also re-adds essential and nice-to-have requirements to the participants lot, as all lots need to have these - without them there is no way to tell who is eligible for a brief and no way to discriminate between supplier responses.


robyoung:
:+1:

PR #214, opened 2016-04-01
Criteria weighting questions

allait:
### Add brief criteria weighting questions
Adds question files for technical, cultural and price weightings for three DOS lots. Questions are split between outcomes & specialists and participants lots, since allowed value ranges are different.

"Cultural fit" is replaced with "Availability" for the participants lot, but the field name is the same to simplify the API validation logic.

### Add `limits` key definition to the question schema
`limits` is a way to group together validation-related values

### Add new total_is_100 validation to the questions schema
`total_is_100` error is returned when criteria weightings don't add up to a 100%

### Use limits when generating percentage question schema
Possible limits keys can change minimum and maximum field values and change the field type to integer so that it doesn't accept decimal places.


README.md
```diff
@@ -32,6 +32,10 @@ Question keys
     * `"on"` service data key name to use for comparison (e.g. "lot" for lots)
     * `being` a list of acceptable values for the key. If service data key value matches one of the values in the `being` the
        question is kept, otherwise the question is removed from the section for the given service
+* `limits` a dictionary of validation/type requirements to limit the possible question values. Available fields:
+    * `min_value` integer, sets the minimum value for percentage questions
+    * `max_value` integer, sets the maximum value for percentage questions
+    * `integer_only` boolean, when `true` makes the percentage question accept only integer values instead of numbers
```
robyoung: `limits` seems really generic but then the description for `min_value` seems really specific. I'm unsure whether the intention is to use `limits` for other types as well (in which case `min_value` seems a bit vague) or just for the percentage type (in which case `limits` seems a bit vague).

schema_generator/__init__.py
```diff
@@ -238,10 +238,10 @@ def pricing_property(question):
...
 def percentage_property(question):
     return {question['id']: {
-        "exclusiveMaximum": True,
-        "maximum": 100,
-        "minimum": 0,
-        "type": "number"
+        "exclusiveMaximum": not question.get('limits', {}).get('integer_only'),
+        "maximum": question.get('limits', {}).get('max_value') or 100,
+        "minimum": question.get('limits', {}).get('min_value') or 0,
+        "type": "integer" if question.get('limits', {}).get('integer_only') else "number"
```
robyoung: This could do with a few tests.

carolinegreen:
Really minor point, but maybe total_is_100 would make for sense if was called total_is_not_100 or total_should_be_100 as this would fit with the other error names following the convention of the name reflecting what something should be or is not.

allait:
@carolinegreen it does fit with the most other validator names: `answer_required,` `under_10_words` etc all describe the expected result rather than the error. This doesn't match how API treats these keys, since they're returned in the `{"errors": ...}` dictionary, which makes it a bit confusing. But I'm sticking with the existing convention for now until we have time to change all `validators` into `errors`.

README.md
allait: The idea behind `limits` is to group all validation-related fields together. These are used primarily by the schema generation and can apply to questions of all types. I'm not attached to this nested structure since it doesn't add that much in the end, but it feels a bit cleaner than mixing all the fields in one question namespace.

I'm not sure what you mean by "in which case min_value seems a bit vague"?

Limit fields themselves could be type-specific or more general, so if `min_value` makes sense for other field types it should be fine to reuse it. The documentation would have to be updated if that happens.

carolinegreen:
@allait  'answer_required' means it needs an answer.  If it was the same it would just be 'answer' on it's own.  That's what I meant by it being a description of something that should be there.  The other name before it is 'not_a_number', likewise the equivalent here would be 'is_a_number'.  It's a subtle difference and a minor point though.

allait:
@carolinegreen all the old ones describe the requirement: `under_10_words`, `file_is_less_than_5mb`, `under_character_limit` etc. But it looks like the newer keys (`max_less_than_min`, `invalid_format`) describe the actual error, so you might be right. I'll rename it to `total_should_be_100`, it makes more sense in the API and validators are a mix already anyway. Good call!

schema_generator/__init__.py
allait: Added some tests

robyoung:
:+1:

PR #218, opened 2016-04-07
G cloud 8 supplier declaration - Cath training with Tom

cathrooney:
Initial changes for the G-Cloud 8 supplier declaration made during training with Tom


TheDoubleK:
Looks good :+1:

PR #219, opened 2016-04-07
Make tests better

tombye:
A PEP8 error with a file from this repo was breaking the tests on the buyer app.

This adds PEP8 checking here to ensure this won't happen again and wraps it in a `Makefile` to make it consistent with the other apps.


allait:
:+1:, but buyer app probably should exclude these files from its own pep8 run [like this](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/setup.cfg#L2)

PR #220, opened 2016-04-13
Mirror sections and (section) descriptions in the prototype

pcraig3:
This pull request:

- Creates the sections [as they currently are in the prototype](http://dm-prototype.herokuapp.com/outcome-overview).
- Adds descriptions to the sections as they currently read in the prototype.  Using the `description` key instead of the `summary_page_description` key because we saw it first.  Open to changing it if anyone has strong opinions.
- Adds a `step` key to manifest files so that we can group sections more precisely on the overview page.
- Adds an extra check for a multiquestion's `any_of` field before generating all of our crazy `anyOf`s and `dependencies` in JSON schemas.

It doesn't:
- Modify any existing questions.  In creating new sections, I roughly grouped questions where they might make sense, but generally the questions in the prototype are quite different.
- Add dynamic content to descriptions.  [There's a section description in the prototype that wants the current user's email address](http://dm-prototype.herokuapp.com/outcomesqa), but we can't really do this with our static descriptions.


allait:
👍

PR #221, opened 2016-04-13
Add question and answer session details question

allait:
Adds the question file and a manifest section. The section name
is temporary since the manifest will have to be changed when the
overview page is done.


pcraig3:
:+1: 

:boom:

PR #222, opened 2016-04-15
Set title as the first question needed for creating a brief

pcraig3:
Previously, our assumption had been that titles were of lesser importance/value than actually picking a location or role (arguably, it gets the buyer thinking of the important thing about their brief sooner), but now we don't know if that's a real thing.
@ralph-hawkins isn't here and the prototype starts with selecting a title, so let's do that.


TheDoubleK:
:+1:

PR #223, opened 2016-04-15
Bump version to 0.25.1

pcraig3:
Reorders the brief manifest so that the title is the first question for creating any brief.


tombye:
:+1:

PR #224, opened 2016-04-15
Fix brief response email and pricing question types

allait:
### Remove unsupported 'email' question type
Adding a new question type requires adding matching toolkit templates and application Jinja macros used to render section forms. Since none of them exist for 'email' question type it raises an exception when the question form is rendered.

The only difference betweeen 'email' and 'text' questions is the additional validation pattern added to the JSON schema. So instead of creating a separate type we can add a new 'limits' property to set question schema 'format'.

### Make brief response dayRate a single-field pricing question
Adds 'price' field support to JSON schema generator and changes dayRate question type to 'pricing'.

This adds validation rules to the API schemas that check that the form field only accepts numeric values and allows us to use error values that we use for other pricing fields.


TheDoubleK:
:+1:

PR #225, opened 2016-04-15
[#116711753] Content update: Title, location and description of work 

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/116711753

This updates the brief questions for outcomes and suppliers to match the content in the prototype (with exceptions as detailed in the story notes).

The question ordering for specialists is necessarily slightly different from the prototype because some questions common to both outcomes and specialists lots are in different orders in the prototype.  It is currently not possible to model this in our manifest file.

I do not think this needs a version bump as it isn't a version that should be used in any app - there are more content changes to come very soon, and this can't really be viewed in the frontend properly until the "grandparent page" story is ready and merged.   So it also isn't worth re-generating the schemas at this point either.


frameworks/digital-outcomes-and-specialists/questions/briefs/workplace.yml
```diff
@@ -1,6 +1,6 @@
-name: Important dates
-question: Any important dates
-hint: 'eg a policy date by which the minimum viable product must be ready'
+name: Address where the work will take place
```
allait: Question name should probably reference the fact that it's supposed to be an address

schemas/questions.json
```diff
@@ -24,6 +24,9 @@
         "pricing", "upload", "percentage", "textbox_large", "multiquestion", "email"
       ]
     },
+    "question_advice": {
```
allait: We should add a description of the new field to README

allait:
👍

PR #227, opened 2016-04-19
Update brief question limits

allait:
### Parse validation errors for list question items
Allows setting character and word length limits for list items.
Adds a flag to parse_quesiton_limits to ignore question.optional,
since we don't need to allow empty strings for list questions.

### Add `number_of_items` field to list questions
Allows setting the maximum number of list items. Needs to be
connected to the application macros rendering the list form.

### Rename under_10_items validator to max_items_limit
Items limit validator used to always have the same limit of 10 items,
which means that API always returns `under_10_items` error. Now that
the item limit is configurable we'd have to generate the correct
error message in the API (like we do with 'under_N_words' errors).

Renaming the validator so the name makes sense for any given limit
we can avoid having to parse validation rules in the API to always
return the same error key, which is replaced with a right message
in the question content.

This also removes the needless duplication with the number of items
field, validator name and error message all containing the same number.


TheDoubleK:
:+1:

PR #228, opened 2016-04-19
Use markdown instead of HTML markup for abbreviations

pcraig3:
Abbreviations are supported in a markup extension which [we have been enabled](https://github.com/alphagov/digitalmarketplace-utils/pull/257).

Also change `>`s to `|` for longer blocks in our messages.
Means we can cut out a bunch of superfluous linebreaks.


TheDoubleK:
:+1:

PR #234, opened 2016-04-20
Reword question for number of suppliers

ralph-hawkins:
change 'How many suppliers/specialists you want to evaluate' to 'How many suppliers/specialists will be evaluated' so the question makes sense in the opportunity page and is consistent with how we asked questions elsewhere.


allait:
👍

PR #239, opened 2016-04-21
content update so buyers know when they should hold their q+a session

ralph-hawkins:
the line 'the date and time of the session' has had '(this should happen within 1 week of publishing your requirements)' added to it.


tombye:
👍

PR #241, opened 2016-04-21
Fix broken output of brief responses

TheDoubleK:
Restores the `supplierName` question for outputting responses that I removed in haste by mistake yesterday.


tombye:
👍

PR #248, opened 2016-04-22
Change "to an" to "for this"

pcraig3:
This is a fallback title no one should ever actually see, but it's still worth keeping it consistent from the way we talk about applying to opportunities in the buyer app.


ralph-hawkins:
👍

PR #251, opened 2016-04-25
Add lead-in text to nice-to-have and essential requirements

allait:
Adds "The supplier must" and "The supplier can" bold lead-in text to the question advice.

The text will appear directly above the list input as long as the question doesn't have `hint` set. With the current template `hint` will appear between `question_advice` and the input.

This is a less disruptive change than adding a new quesiton feature to the templates and apps and it allows us to avoid what feels like a very rarely used question field.

screen shot 2016-04-25 at 10 38 08

screen shot 2016-04-25 at 10 38 24


carolinegreen:
👍

PR #253, opened 2016-04-25
Content for opportunities

ralph-hawkins:
added 'Buyers will use the essential and nice-to-have skills and experience to help them evaluate suppliers.' section description and changed 'Evalutaion' heading to 'How suppliers will be evaluated


allait:
needs a version bump, other than that 👍

PR #254, opened 2016-04-25
Updates to 'capped time and materials' description

tombye:
As requested in this story: https://www.pivotaltracker.com/story/show/118286875

### DIFF
--- a/frameworks/digital-outcomes-and-specialists/questions/briefs/priceCriteria.yml
+++ b/frameworks/digital-outcomes-and-specialists/questions/briefs/priceCriteria.yml
@@ -19,11 +19,10 @@ options:
   -
     label: Capped time and materials
     description: >
-      The capped time and materials approach combines elements of the fixed price and time and materials methods. 
-      The supplier does the work on a day rate and expenses basis but you need to agree date which they can’t work 
-      past. Most buyers choose a capped time and materials approach for alpha, beta and live. They also tend to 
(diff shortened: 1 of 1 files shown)


TheDoubleK:
`If the supplier finishes the work before they said they would, you still have to pay the full amount.` - this makes it sound just like a fixed price.

sarahonetui:
The difference is the price is given in day rates and so should have less extra risk (cost) added by the supplier than in a fixed price contract. That said, I agree with you, they're very similar. It's one of the price options in the framework so we have to include it.

TheDoubleK:
Might it make the difference more obvious to say something like `If the supplier finishes the work before they said they would, you only have to pay the amount for the time they have taken.` - or is that not true?
In which case could it be:  `If the supplier finishes the work before they said they would, you still have to pay the full amount specified in their original quote`?

It just feels like it need a few more words to make the difference clearer either way.

sarahonetui:
The first statement isn't true, you have to pay the amount you capped the work at even if they finish early. Agreed with @TheDoubleK that probably best to leave as it is.

TheDoubleK:
:+1:

tombye:
Just to be clear, are these changes ok to be merged or does 'leave it as it is' mean to close this?

sarahonetui:
Sorry. To be clear, we're going with the new content as it was in the story.

PR #257, opened 2016-04-26
Update capped time and materials definition

allait:
#### [#118374227] Change 'availability' column name in supplier responses csv

Changing the question name to match the new question text.
Question name is not displayed anywhere except for the output CSV.

#### [#118292307] Fix description typo in backgroundInformation

Bump version to 0.32.1


TheDoubleK:
:+1:

PR #260, opened 2016-04-26
Research dates contraction

ralph-hawkins:
Guidance now says 'If you *don’t* have exact dates, say when you expect research to start and end.'


allait:
👍

PR #261, opened 2016-04-26
How youll assess suppliers

ralph-hawkins:
Used to be 'how you'll assess the suppliers' now its 'how you'll assess suppliers'.


allait:
👍

PR #263, opened 2016-04-26
Last minute content update

ralph-hawkins:
- Change 'proposed solution' to 'proposal everywhere
- Added 'to' to 'Try not take too many...'
- Explain availability on participants weighting
- Fixed some dumb quotes to apostrophes


allait:
👍 1.0 🎆

PR #264, opened 2016-04-27
Add G-Cloud 8 Messages

carolinegreen:
Add G-Cloud 8 messages.  Removed template options as framework notices are being deprecated.  Leaving G-Cloud 8 as plaintext and not an html entity as I think it would be better to handle this in the app.  This leaves the yml more readable and purely for content as originally intended.


TheDoubleK:
I think if we're going to do HTML entities at all then they need to go here in the content, I can't see how the front-end apps can know how to swap them in? Not without doing something really ugly and brittle, anyway.

But we can investigate that as part of the entities story/bug here: https://www.pivotaltracker.com/story/show/118404225

:+1: for this.

carolinegreen:
I think it would be less brittle to do a regex than relying on people to put in html entities each time.  This leaves less room for errors.  

Also if the whole point of separating things into yml is to make things easier for everyone to add content, then I'm not sure forcing people to have to use html entities, which they shouldn't necessarily have to take into consideration in this context, helps with this.

TheDoubleK:
But, given that the front-end apps shouldn't know about what frameworks do or don't exist, what regex could you possibly use?  We just made all the quotes in the content curly; if we want HTML entities then it's the same thing as the curly quotes.

carolinegreen:
The content loader searches for frameworks and does know that they exist.  While that's outside of the frontend apps, it also isn't in the content itself.  It can't be that html entities all have to be entered manually.

PR #266, opened 2016-05-04
Make numberOfSuppliers brief question a number question

allait:
### Rename percentage question type to 'number'
Allows us to generalise the percentage type to use it for all number fields. The main difference between number and text fields is the type of the value returned, so percentage fields are a subtype of number fields with unit set to '%' and max value set to 100.

### Change 'percentage' questions to 'number' type
Adds unit, unit_in_full and unit_positition fields to make sure percentage fields are rendered correctly. Since we're keeping the existing percentage defaults for min and max value (0 and 100), there's no need to add limits to the questions that didn't define them right now.

### Require `unit_` fields to be set if `unit` is set
Also restricts `unit` field to 'number' questions using JSON schema dependencies.

### Make numberOfSuppliers brief question a number question
Adds min and max value limits, validation message and stores the value as integer in the database.


TheDoubleK:
Looks good - the `exclusiveMaximum` logic confused me for a bit, but it does the right thing :+1:

PR #267, opened 2016-05-06
Raise the number of suppliers to be evaluated max to 15 

allait:
Matches the max value in the published briefs.

Bump version to 1.3.1


pcraig3:
👍

PR #269, opened 2016-05-10
Don't markdown dates

TheDoubleK:
Required to fix:
 - https://www.pivotaltracker.com/story/show/119086793
 - https://www.pivotaltracker.com/story/show/119075005

The markdown filter wraps values in `<p>` tags but we generally want to use dates inline and the `<p>` breaks the page layouts in the frontend.  So we're not going to use markdown for date strings.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists/messages/dates.yml
+++ b/frameworks/digital-outcomes-and-specialists/messages/dates.yml
@@ -1,14 +1,9 @@
-clarifications_close_date: |
-  5pm GMT, 6 January 2016
+clarifications_close_date: '5pm GMT, 6 January 2016'
 
-  *[GMT]: Greenwich Mean Time
-clarifications_publish_date: |
-  5pm GMT, 13 January 2016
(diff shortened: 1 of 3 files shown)


allait:
This is removing the annotation lists and I think there was a reason why we added them. @pcraig3 probably knows more about this.

TheDoubleK:
I discussed this with @pcraig3 earlier.  The markdown filter created abbreviation tags that expand "BST" into "British Summer Time", which would be shown if the user knew to hover their mouse over the "BST" on the page.

But as markdown breaks things the alternative way to do it is to have the `<abbr>` tags in the content here, and `|safe` all dates in the frontend apps.  
We decided that the number of cases where the user:
 * doesn't understand `BST` *and*
 * knows they might be able to hover their mouse over `BST` to find out what it means
are probably vanishingly small, so we were happy to just use strings for dates and not have to `|safe` in the frontend.

pcraig3:
I'm disappointed, but I agree with Kev.

:disappointed::+1:

PR #271, opened 2016-05-11
Add guidance on only asking for a single specialist

ralph-hawkins:


TheDoubleK:
Version now needs to go to 1.3.5

TheDoubleK:
Participants and outcomes versions look identical?  Or did I miss something.  There's no need to split them out if the words are all the same.

pcraig3:
Looks 👍  to me.

PR #273, opened 2016-05-12
G-Cloud 8 - small updates

cathrooney:
G-Cloud 8 changes:
 - Add DUNs number question to About us on supplier declaration

Following feedback from Eleanor:
 - Remove redundant option from Networks question
 - Update PSN options in data in transit protection
- Add Microsoft Edge as a browser choice


frameworks/g-cloud-8/questions/declaration/dunsNumber.yml
```diff
@@ -0,0 +1,3 @@
+question: What is your organisation’s DUNs number?
```
TheDoubleK: Everywhere else we have all capitals - i.e. `DUNS` rather than `DUNs`.  We should make it the same everywhere.

TheDoubleK:
I've rebased this and bumped the version number up, but let's wait for final content for the DUNS number question before merging.

PR #277, opened 2016-05-13
changes proposed solution to proposal criteria

ralph-hawkins:
**old**

Technical comptence: proposed solution criteria

**new**

Technical competence: proposal evaluation criteria

- we want to refer to proposal instead of proposed solution and spell competence correctly.


carolinegreen:
👍

PR #278, opened 2016-05-16
G cloud 8 final changes

cathrooney:
Update ZIP file size from 3.2MB to 1.2MB in five declaration questions
Tweaked DUNS number copy with Roz


TheDoubleK:
Waiting on G8 supplier pack file to be ready so we can update this with the correct file size in the download links.

PR #279, opened 2016-05-16
Missed one PSN

TheDoubleK:
This should have been updated with the others but got missed.

c.f. https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/g-cloud-8/questions/services/dataProtectionBetweenUserAndService.yml#L16


allait:
👍

PR #280, opened 2016-05-19
Re add zip file size to download links

TheDoubleK:
We removed the file size before G8 launch because we only got the supplier pack at the last minute and didn't know the file size.

This re-adds the correct file size, so people can tell if they want to click the links or not.


allait:
👍

PR #283, opened 2016-06-03
Mandatory assessment methods

allait:
### Add `before_summary_value` question key
Required for [#118392107](https://www.pivotaltracker.com/story/show/118392107). To add "mandatory" options to the brief view we need a way to add additional data to question summary value. So we're adding a new key to store this. Values from `before_summary_value` are never sent to the API, they're only used by the content loader for display in the summary tables.

### Make assessment methods quesiton optional
Since there's a mandatory value that's always added the question itself is now optional.


frameworks/digital-outcomes-and-specialists/questions/briefs/evaluationTypeOutcomes.yml
```diff
@@ -2,20 +2,20 @@ name: Assessment methods
 id: evaluationType
 question: How you’ll assess suppliers
 question_advice: |
-  Choose how you want to assess suppliers.
-  
-  You can only use the methods you choose here.
-  
-  You must ask for a written proposal.
-  
+  All suppliers that reach the assessment stage will be asked to provide a written proposal.
```
TheDoubleK: The wording here (and for other lots) implies that a written proposal/work history will be asked for automatically by the system.  Is that true? (I don't think so?)

Would something like `You must ask all suppliers that reach the assessment stage for a written proposal.` be better, to make it clear that it's something they have to do, not something we will do for them.

TheDoubleK: Maybe one for @superroz?

allait: Is this a quote from @superroz?

TheDoubleK: Yes, I think Roz must have edited my comment.  I didn't type the second paragraph.

TheDoubleK:
Let's do it :+1:

PR #285, opened 2016-06-13
Reword maximum suppliers

TheDoubleK:
This is shown on the public view of an opportunity, where the previous wording ("how many you''ll evaluate") didn't make sense.  This way it works both as a question for the buyer when writing their requirements and for the suppliers viewing the opportunity.


TheDoubleK:
@ralph-hawkins says :+1: in real life.

ralph-hawkins:
👍

PR #286, opened 2016-06-16
Update content on specialist budget page

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/121309131) story on Pivotal.

This PR adds an extra line of content to the budgetRangeSpecialists.yml which is used in the buyer front end to reinforce the message that if a buyer wants to shortlist on day rate, they have to provide a budget day rate. 

It also bumps the framework version to 1.5.3


pcraig3:
👍

PR #292, opened 2016-07-04
Create script to extract validation messages

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/121311097) Pivotal story.

This script will extract the validation messages for all questions that have them, from the frameworks yaml files, and save them in a new directory in an easy to edit yaml file for each separate framework.

Validation messages are saved with their corresponding question, hint, name and the path to the file they are in to allow easy recognition of the messages context.

There are a few more validation messages contained within the front end apps which will need to be extracted by hand.

See below for usage example.

double 3


TheDoubleK:
Has this been run already and results passed on at all?

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+
+import yaml, os, shutil
+
+dir = os.path.dirname(__file__)
+
+if os.path.exists('{}/validation_messages'.format(dir)):
+    shutil.rmtree('{}/validation_messages'.format(dir))
```
allait: This feels a bit dangerous. Should we warn that the folder exists and exit instead? Or maybe ask for the path to save files to.

Wynndow:
@TheDoubleK It's been run, but the results not passed on. Was waiting to get some eyes on it in case it was doing anything horrendous I wasn't aware of!

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+
+import yaml, os, shutil
+
+dir = os.path.dirname(__file__)
```
risicle: Not tremendously important, but `dir` is a builtin in python and it's best to avoid reassigning builtin names to something else. (The most common culprit is `set`)

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
```
risicle: Would be nice to have a docstring for this file which describes exactly what it does and how to use it.

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,51 @@
...
+
+import yaml, os, shutil
+
+dir = os.path.dirname(__file__)
+
+if os.path.exists('{}/validation_messages'.format(dir)):
+    shutil.rmtree('{}/validation_messages'.format(dir))
+
+os.makedirs('{}/validation_messages'.format(dir))
```
risicle: All this code which actually *does* something (rather than just defines things) feels like it should belong in the `if __name__ == "__main__"` section. Otherwise simply importing this module from another (for some reason) will cause all sorts of filesystem altering events to happen.

scripts/extract_validation_messages.py
Wynndow: @allait Have updated to use docopt to take an output path as an argument, and also to abort the script if the output already exists.

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,51 @@
...
+        with open(file_path, 'r') as f:
+            doc = yaml.load(f)
+            if doc.get('validations'):
+                data = create_data_packet(file_path, doc)
+                with open (output_file_name, 'a') as output_file:
+                    write_back_to_file(output_file, data)
+
+
+if __name__ == '__main__':
+    for dirName, subdirList, fileList in os.walk('{}/../frameworks'.format(dir)):
```
risicle: (general point)  "Well behaved" python "scripts" tend to follow a pattern where it's just as easy to import the module from some other python and run a function to perform the action as it is to run it from the command-line. So a really neat thing to do is implement the whole _meat_ of the script in a _function_ (which maybe takes a few args) and make the contents of the `if __name__ == "__main__"` section simply call that function with some default arguments (or if you're accepting command-line arguments, process those arguments and pass them through to the function call)

This way if I wanted to do a validation message extraction from within another script I might be able to do e.g.

```
from extract_validation_messages import extract_validation_messages

extract_validation_messages(out_dir="/some/special/dir", frameworks_dir="/another/specific/dir")
```

scripts/extract_validation_messages.py
Wynndow: @risicle Yep, just moved it all down, as well as renaming dir to something more better.

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,67 @@
...
+        output.append(with_paths)
+    return [val for sublist in output for val in sublist]
+
+def read_and_write_files(file_paths, output_file_name):
+    for file_path in file_paths:
+        with open(file_path, 'r') as f:
+            doc = yaml.load(f)
+            if doc.get('validations'):
+                data = create_data_packet(file_path, doc)
+                with open (output_file_name, 'a') as output_file:
```
risicle: You don't need to reopen the file for each line you're going to add, could just put this once at the toplevel of the function.

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,67 @@
...
+def check_and_create_output_directory(path):
+    if os.path.exists('{}/validation_messages'.format(path)):
+        print('{}/validation_messages already exists. Please specify a different output path.'.format(path))
+        sys.exit()
+
+    os.makedirs('{}/validation_messages'.format(path))
+
+def create_data_packet(file_path, doc):
+    return dict(
+        path=file_path.split('..')[-1],
```
risicle: Can see why you're doing this but... `bit odd` to be specifically depending on the `../` being in the path in the right place. Just a bit fragile to future adaptations (no, I don't have a suggested alternative...)

risicle:
Soft-thumbs-up from me (read: I don't think I should have the authority to use the sacred emoji yet)

scripts/extract_validation_messages.py
Wynndow: Yep, makes much more sense. Have changed.

scripts/extract_validation_messages.py
Wynndow: Yeah, it's because of the relative nature of the script to the files in question. Could be overcome by passing in the path to the files as an argument? Not sure if it's worth it for this though?

scripts/extract_validation_messages.py
```diff
@@ -0,0 +1,67 @@
...
+    extract_validation_messages.py --output-path=<output_path>
+
+"""
+import yaml, os, shutil, sys
+from docopt import docopt
+
+def check_and_create_output_directory(path):
+    if os.path.exists('{}/validation_messages'.format(path)):
+        print('{}/validation_messages already exists. Please specify a different output path.'.format(path))
+        sys.exit()
```
allait: Since we're aborting the script we could use `sys.exit(1)` to give it a non-zero exit code in case we ever use it in Jenkins/Travis.

allait:
👍

PR #293, opened 2016-07-20
Add per-lot description support for section descriptions  …

allait:
We need a way to show a different description based on the brief response lot. We've looked at possible approaches (using a different `lot_description`, not allowing a default option and allowing to filter on different fields apart from by lot) and decided to allow setting description to a dictionary of `<lot_slug>: <description_for_lot>` pairs with an optional special `default` key to avoid duplicating matching descriptions.

We're adding a list of existing lots to the schemas to make sure any typos are caught by YAML tests. This is a flat list of all existing lots for now, since the schemas are shared between different frameworks.


scripts/extract_validation_messages.py
```diff
@@ -62,6 +71,7 @@ def read_and_write_files(file_paths, output_file_name):
     check_and_create_output_directory(arguments['--output-path'])
     for dirName, subdirList, fileList in os.walk('{}/../frameworks'.format(current_dir)):
         if directory_is_questions_(dirName):
-            output_file_name = '{}/validation_messages/{}.yml'.format(arguments['--output-path'], dirName.split('/')[-2])
+            output_file_name = '{}/validation_messages/{}.yml'.format(arguments['--output-path'],
```
TheDoubleK: How did we live with this enormous line in the past?!? Thank the Lord for pep8.

TheDoubleK:
Nicely done :+1:

PR #295, opened 2016-08-01
Open not live requirements

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal.

Briefs should be described as being 'open' and not 'live'.

This PR is just a simple change of content and version bump.


TheDoubleK:
:+1:

PR #296, opened 2016-08-01
Open, not live

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal.

This PR is to change the content to refer to briefs as 'open' instead of 'live'. Very similar to the last one of this story, I missed this occurrence of 'live' in the last one.


TheDoubleK:
:+1:

PR #299, opened 2016-08-17
[#127807161] Add G-Cloud 8 contract variation

TheDoubleK:
This is the hopefully "final" version of the contract variation content for this story: https://www.pivotaltracker.com/story/show/127807161

The inserts and deletes are marked up with HTML here in the content. Not the prettiest, but I can't see a reasonable way round it.

Page content from Andy, clause amendments from Alexa.


idavidmcdonald:
👍 💯 
Nice. I didn't know about `<del>` and `<ins>` tags before.

PR #304, opened 2016-08-31
Small update to change h1 tags to h2 tags

pcraig3:
`# this is an h1`

# this is an h1

`## this is an h2`

## this is an h2

***

Too many `h1` tags on the page are breaking our functional tests, as well as being a semantic no-no.

Small update makes them `h2`s, [which is what we're doing in other places](https://github.com/alphagov/digitalmarketplace-frameworks/blob/344b6699648a2a2974ebaed2ebe0b309d4e21866/frameworks/digital-outcomes-and-specialists/manifests/edit_brief.yml#L66).  Should fix the functional tests.


ralph-hawkins:
👍

PR #305, opened 2016-09-05
Remove G-Cloud 8 URLs file

TheDoubleK:
It's been over a month and it seems like CCS aren't going to make the G-Cloud 8 framework page on their website any time soon.

By removing the URL file here it means we can get the 33 day old pull-request in without adding broken links to our app.


idavidmcdonald:
👍

PR #307, opened 2016-09-16
Things that participant suppliers have already agreed to provide

ralph-hawkins:
Splits out the terms and conditions page for Participants lot.

Adds the following to the content:
The suppliers have already agreed that:
- their prices must include all incentives, recruitment, and travel and subsistence costs paid to participants
- you don’t pay for participants who don’t attend on the day
- they are responsible for paying participants directly

Looks like this:
screen shot 2016-09-16 at 12 09 29


TheDoubleK:
Good stuff :+1:

PR #309, opened 2016-09-29
[#128669753] Split description for before and after agreeing G8 variation

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/128669753

The introductory paragraph needs to be different at beginning and end, so split into two separate complete sections rather than just having an "extra" bit.

I've added a major version bump as this is incompatible with the old format.


idavidmcdonald:
👍

PR #311, opened 2016-09-30
DOS2 supplier declaration - ESPD tweaks + provisional dates added

cathrooney:
I have made the small number of changes to align our mandatory questions to the EU ESPD.

While I was in there, I updated the provisional application dates


frameworks/digital-outcomes-and-specialists-2/manifests/declaration.yml
```diff
@@ -43,7 +43,8 @@
     Check that your organisation complies with [Regulation 57(1) of the Public
     Contracts Regulations 2015](http://www.legislation.gov.uk/uksi/2015/102/regulation/57/made).
   questions:
-    - conspiracyCorruptionBribery
+    - conspiracy
+    - CorruptionBribery
```
allait: We should keep this and the filename as `corruptionBribery` to avoid any issues when switching between case-insensitive and case-sensitive file systems

TheDoubleK:
This looks good to me now.  I've rebased it, probably fine to merge it without a version bump, as there are more frameworks changes coming before DOS2 gets switched on (e.g. splitting the address).

PR #312, opened 2016-10-04
Split admin service edit manifests

allait:
### [#131440081] Allow admin to edit Support section for G-Cloud 7 & 8
Copies the entire section from the service submission to keep the list of questions consistent between the manifests.

Not adding this for G6 since G6 services are no longer publicly visible in the catalogue.

### Add admin service edit manifest for DOS and DOS 2
The admin app was using g-cloud-6 manifest for all service editing, but thanks to a number of coincidences the manifest worked fine with DOS services:

* Because all g-cloud-6 questions define an explicit list of lots, all sections are filtered out for DOS services
* Since G6 defines `lotName` and `id` questions that match DOS service keys those fields are displayed on the page even though DOS framework content doesn't have these questions

Changing the admin app to use a framework specific manifest when editing a service requires these manifest files to exist for DOS.

Adding a simple manifest that only allows editing service titles. Since DOS content doesn't have `id` or `lotName` questions at the moment, removing the "Service attributes" section that displayed them on the admin service page.


frameworks/g-cloud-7/manifests/edit_service_as_admin.yml
```diff
@@ -35,3 +35,12 @@
     - termsAndConditionsDocumentURL
     - sfiaRateDocumentURL
     - pricingDocumentURL
+-
+  name: Support
+  editable: True
```
pcraig3: Should this be lowercase? 
It doesn't seem to matter to our apps, but we've always written it as lowercase everywhere else.

frameworks/g-cloud-8/manifests/edit_service_as_admin.yml
```diff
@@ -35,3 +35,12 @@
     - termsAndConditionsDocumentURL
     - sfiaRateDocumentURL
     - pricingDocumentURL
+-
+  name: Support
+  editable: True
```
pcraig3: lowercase?

PR #313, opened 2016-10-04
DOS2 remove zip links

cathrooney:
Removing links to the ZIP file and pointing to the overview page instead


TheDoubleK:
Might be worth parking this until the files story is played and we know the URLs (they are likely to be `assets.digitalmarketplace...` URLs rather than `/suppliers`).  At that point we can rebase and bump the version number.

PR #316, opened 2016-10-17
Generate DOS2 briefs and responses schemas

TheDoubleK:
We want the script to generate schemas for briefs as well as services.


idavidmcdonald:
👍

PR #317, opened 2016-10-20
[#132662099] Use raw unicode not HTML escapes

TheDoubleK:
First part of a fix for this bug: https://www.pivotaltracker.com/story/show/132662099

**This is up for discussion... is it the way to go?**

I've tested pulling this content into both buyer and supplier apps and removing some relevant `|safe` filters and everything looks good still as far as I can see - non-breaking spaces and hyphens don't break and horizontal ellipsis looks right on the page.

PROS:
- We will be able to get rid of some of the `|safe` from our Jinja templates (both toolkit and apps)

CONS: 
- It isn't at all obvious to people editing the content which characters are "special", and people adding new content might not remember or even know how to add the "special" characters.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-2/manifests/declaration.yml
+++ b/frameworks/digital-outcomes-and-specialists-2/manifests/declaration.yml
@@ -3,7 +3,7 @@
   editable: True
   description: >
     You must make this declaration to apply to Digital Outcomes and
-    Specialists&nbsp;2. You can come back and change your answers at any time before
+    Specialists 2. You can come back and change your answers at any time before
     5pm, 6 December 2016.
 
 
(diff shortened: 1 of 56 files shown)


andymcaleer:
If it's safer/better to use unicode, I'm sure content designers can learn this. We can create a cribsheet, say, that includes this advice for all in-app work (or maybe add it to the YAML markup guidance on the alphagov/digitalmarketplace-frameworks homepage).

TheDoubleK:
OK great, I've removed the "FOR DISCUSSION"from the title and this now wants a review and hopefully a thumbs-up.

risicle:
Are you sure we should use _raw_ escapes? I would favour something like unicode escapes, because so many of the unicode characters are either invisible or visually indistinguishable from regular characters.

To us developers who are intimately familiar with our text editors, this may be fine, but the principal editors of this repo will be (?) non-developers, possibly pasting content from a word processor...

There is a slight complication in that yaml seems to require strings including `\uNNNN` escapes to be `"double-quote wrapped"`, which, I think, interferes with the entry of multi-line strings...?

TheDoubleK:
Closing this as now HTML is going to be OK in content.

PR #322, opened 2016-10-25
Use jinja templating instead of nested descriptions

pcraig3:
Including jinja logic in our content means we can totally scrap the logic we had around putting different descriptions in nested keys.


pcraig3:
Closing this PR until we're ready to merge handling variables in the content loader.

PR #323, opened 2016-10-26
More DOS2 content updates

TheDoubleK:
This:
 1. Changes the word "region" to "location" everywhere in the supplier application content for consistency
 2. Updates the "view locations" link to a different PDF file because the old link is now flagged as insecure by browsers
 3. Fixes the "conspiracy" question to not use a nested list of bullets


andrewchong:
Looks good to me. 👍

PR #324, opened 2016-10-27
[#131617119]add supplier_guide_url messages for all frameworks

risicle:
Story https://www.pivotaltracker.com/story/show/131617119

@TheDoubleK So I guess we probably should add this file for all the other frameworks. For all the g-cloud ones this appears to be https://www.gov.uk/guidance/g-cloud-suppliers-guide#how-to-apply and I'm guessing it's the same url as this for DOS1. Which of course makes me think it's metaframework-based, and so we should possibly be driving it from that. But seeing as the anchors seem to be different for the different metaframeworks makes me think it could never be straightforward enough to allow us to do that.


risicle:
Scratch original comment - I've switched things around & added `supplier_guide_url` for all frameworks

TheDoubleK:
:+1:

PR #329, opened 2016-11-09
Fix trailing quotes

allait:
Converting markdown questions to multi-line strings left some trailing quotes that we need to remove.


pcraig3:
👍  for real this time.

PR #330, opened 2016-11-15
Legacy edit brief response manifest

idavidmcdonald:
We will be running two different manifests to edit a brief response for a period of two weeks. We therefore make the current manifest a legacy file. After two weeks this legacy manifest can then be removed. This will leave us with just one manifest which will have been nicely named `edit_brief_response`.

We've also introduced a change log because there wasn't one.


galund:
Don't know if this is a terrible idea, but throwing it out there anyway: another idea might be to have a 'version' number in the json somewhere, indicating which flow / version-of-flow is in use?

The trouble with introducing 'legacy-this' and 'legacy-that' is that in the end, the new ones become legacy too, and then you get a problem :-D

idavidmcdonald:
George, good point. We had a discussion now about the plan forwards with this. We will be deprecating and removing the legacy manifests and legacy questions after a two week period after this launches. Because we can delete them we won't need to maintain multiple versions at the same and think we can go with "legacy" for the moment. We did discuss a versioning alternative but looks like we will be able to not have to do so.

idavidmcdonald:
Currently paused on this pull request until we figure out some later steps in the process for handling multiple flows for applying for brief responses.

tombye:
👍  (also: thanks @galund for the input.)

PR #331, opened 2016-11-16
Add dynamic list and followup questions to schema generator

allait:
Generates an array property for the dynamic list field and sets the items schema to match nested question fields.

Followup questions generate a subschema to validate that if a question was answered with 'Yes' then the followup question has to be answered too.

Schema generation supports multiple followup relationships within one group of dynamic_list nested questions, but it's likely that API error reporting will require some changes if this ever becomes a feature we need.


pcraig3:
Seems like you need to rebase, but 👍  I think.

PR #332, opened 2016-11-17
Updated 'view region' to 'view location' and fixed links.

danielknell:
Text changes to bring things inline with #323 and fixes the security issue with links for DOS questions.

Fixes https://www.pivotaltracker.com/n/projects/1884593/stories/134137657

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-2/questions/briefs/locationOutcomes.yml
+++ b/frameworks/digital-outcomes-and-specialists-2/questions/briefs/locationOutcomes.yml
@@ -6,7 +6,8 @@ question_advice: |
 
   Suppliers have said where they’ll provide outcomes.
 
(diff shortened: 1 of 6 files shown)


allait:
Should we be changing service questions for DOS-1? It's been closed for a while now and there's no way to view those question descriptions now. So the new description doesn't match what all of the suppliers have seen.

TheDoubleK:
Yep I'd be tempted to only update the `/briefs/` questions that are still being displayed to users.  The `/services/` ones it feels like we should leave how they were.

danielknell:
have nuked the commit updating dos entries, and just left the briefs one in place.

TheDoubleK:
:+1: after a rebase. Sorry it took a while to get back to this :(

idavidmcdonald:
It looks like we forgot to include a version bump for this pull request. One can be added retrospectively however by submitted a pull request with just the version bump.

PR #333, opened 2016-11-22
Add `required_value` property to question YAML and schema generator

idavidmcdonald:
We wish to have a boolean question that you can only answer "yes" to, otherwise validation should fail. 

- Add `required_value` property to questions schema for boolean questions
- If a `required_value` is provided for a boolean question, then we wish to add this validation to a generated schema using the `enum` property.


allait:
Very nice 👍 

Let's add the new field to the README and then it's good to go.

PR #334, opened 2016-11-25
Content for new apply for a brief

idavidmcdonald:
We will be running a new process for suppliers applying for a brief. This process will need to run in parallel to the current one for a period of two weeks. After this period we will be able to remove the current process (the `legacy_edit_brief_response` manifest).

### Prototype
https://dm-prototype.cloudapps.digital/structure/opportunity/dos-opportunities?opportunity=0

### Related supplier front end pull request
Coming soon...

### Pivotal stories
Contains content for the following stories:
- https://www.pivotaltracker.com/story/show/129843611
- https://www.pivotaltracker.com/story/show/129839965
- https://www.pivotaltracker.com/story/show/131617615

### Half way house
We are aiming to build the prototype but this pull request only has half of the content done. This is so we can get something small(er) code reviewed/QA/merged etc rather than a massive block of work.

Still to be completed:
- Questions to give evidence for essential requirements, progressive disclosure question for nice to haves with evidence
- Validation messages (some may exist now but all of them will need going over and reviewed at a later stage)

### File names
We have changed filenames for existing questions (i.e. `dayRate` to `legacyDayRate`).
Our new questions can then replace the old questions (i.e. `dayRate`).

Note, some questions filenames have not been migrated (`essentialRequirements` and `niceToHaveRequirements`). This is because the newer versions of these questions may be different question types and therefore named differently meaning no conflicting filenames (e.g. `essentialRequirementsMet`).

The legacy questions are going need to stick around for the foreseeable future as we need to be able to play back information about an old brief forever. We may decide later on to rename these question files to something that is slightly more futureproof.


allait:
I think it might be better to use the un-prefixed names for the new questions and add `old` or `legacy` to the old ones. That way we can remove them once they're no longer needed instead of having to rename the new ones at the same time.

frameworks/digital-outcomes-and-specialists/questions/brief-responses/newAvailabilityOutcomes.yml
```diff
@@ -0,0 +1,19 @@
+name: Earliest start date
+id: availability
+question: When is the earliest the team can start
```
allait: I think this text is the only difference between two questions. We can use the Jinja `{% if lot == ... %}` here and merge them together

idavidmcdonald: This appears to not be possible based on our current URL structure. We can merge the questions into one which is fine. However, then the section that they are used in also needs to be merged into one with jinja in the section name as such:

image

However, the problem with this is that the section name is used in the url and this will then 404 when matching by section name.

image

allait: I think you can define an explicit slug on the section using `slug: ...`

galund: Seems like it might be a good idea if the default slug used the ID, rather than the name?

idavidmcdonald: You can't define a `slug` for a section. You can define a `slug` for a question but this doesn't get passed up to the section.

idavidmcdonald:
@allait In regards to your point using unprefixed names for the new ones and calling the old ones something like 'legacy'.
You mentioned that we can "remove" the old ones when they are no longer needed. It looks like at the moment they will always be needed as we always need to be able to play back brief response information for legacy briefs. Just wanted to check we are on the same page there?
Regardless, I think it does make sense to rename the older ones now and we can decide at a later point if they need to be changed again.

idavidmcdonald:
- Have updated question filenames so older questions are renamed legacy and the new ones can be unprefixed.

- Have not merged the availability question into one using Jinja to decide on the content by lot as this did not appear to be possible as we rely on the section name.

frameworks/digital-outcomes-and-specialists/questions/brief-responses/availabilityOutcomes.yml
```diff
@@ -1,7 +1,8 @@
-name: Date the supplier can start work
+name: Earliest start date
 id: availability
-question: When can the supplier start work?
+question: When is the earliest the team can start
```
Wynndow: I'm on it.

frameworks/digital-outcomes-and-specialists/questions/brief-responses/dayRate.yml
```diff
@@ -1,7 +1,6 @@
 name: Day rate
 question: What’s the specialist’s day rate?
-hint: Enter a number, not a range.
-
+question_advice: "{% if brief.budgetRange is defined %}<h2>Buyer's maximum day rate:</h2><p>{{ brief.budgetRange }}</p>{% endif %}<h2>Your maximum day rate:</h2><p>£{{ max_day_rate }}</p>"
```
allait: I thought we got rid of `is defined`?

allait:
👍

PR #335, opened 2016-11-28
Mention 'bower link' in README as it's very useful

galund:
Does this make sense @allait ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,12 +119,13 @@ Development
 A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
 as follows:
 
-- from this repo, run `bower link`
-- from each app, run `bower link digitalmarketplace-frameworks`
+- from this repo, run `node_modules/bower/bin/bower link`
+- from each app, run `node_modules/bower/bin/bower link digitalmarketplace-frameworks`
 
 Then, whenever the framework content is changed:
 
-- on each frontend app, rebuild the static files to include the new content by running `make frontend_build` (NB this step also required before deployment to PaaS via `cf push`)
+- on each frontend app, rebuild the app's `content` directory to include the new content by running `make frontend_build` (NB this step also required before deployment to PaaS via `cf push`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,6 +108,20 @@ These are the fields that are turned into TemplateFields when the YAML files are
...
+
+A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
+as follows:
+
+- from this repo, run `bower link`
+- from each app, run `bower link digitalmarketplace-frameworks`
+
+Then, whenever the framework content is changed:
+
+- deploy framework content to each frontend app with `make frontend_build` (NB this step required before deployment to PaaS via `cf push`)
```
allait: I think "rebuild" would probably fit better than "deploy". Also I'm not sure if it's clear that you need to run the make step in the app itself, not in the frameworks repo.

allait:
Small not about frontend rebuild, but looks good otherwise 👍

README.md
```diff
@@ -108,6 +108,20 @@ These are the fields that are turned into TemplateFields when the YAML files are
...
+
+A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
+as follows:
+
+- from this repo, run `bower link`
+- from each app, run `bower link digitalmarketplace-frameworks`
+
+Then, whenever the framework content is changed:
+
+- on each frontend app, rebuild the static files to include the new content by running `make frontend_build` (NB this step also required before deployment to PaaS via `cf push`)
```
TheDoubleK: Rather than `make frontend_build` for each change you can instead run `npm run frontend-build:watch` from the relevant frontend app folder, which will re-run the frontend build every time a change is made. (At least, I think this will work for frameworks content changes - it definitely works for template and SCSS changes.)

galund: It doesn't seem to work! As in, the watching bit doesn't work - it does actually copy the updated framework content on first run...

PR #337, opened 2016-11-29
Edited max supplier content

tracy-content:
Changed content following user feedback. Guidance also being updated.


idavidmcdonald:
Is this content only to be updated for DOS2 (which is not currently live so no briefs can be created)? There are yml files for DOS1 that can be changed if we wish to update this for current briefs.

Briefscontent:
Yes DOS1 too... I'll do that!

VERSION.txt
```diff
@@ -1 +1 @@
-5.1.0
\ No newline at end of file
+5.3.0
```
idavidmcdonald: This probably only needs to be a patch version bump rather than a minor version bump. You are probably looking at going to 5.1.2 (assuming Dan takes 5.1.1)

tombye:
@superroz does this need any more looking at?

TheDoubleK:
The commits here could probably be squashed together into one by some kind dev.

superroz:
Looks good to me.

PR #340, opened 2016-12-01
Region to location

danielknell:
Rename the Region fields to Location

https://www.pivotaltracker.com/n/projects/1884593


pcraig3:
If it doesn't, we probably shouldn't be updating `frameworks/digital-outcomes-and-specialists/questions/*.yml` files.

If it does, there's a few that still say "Region" ([for example](https://github.com/alphagov/digitalmarketplace-frameworks/blob/region-to-locatoin/frameworks/digital-outcomes-and-specialists/questions/services/productManagerLocations.yml])).

danielknell:
It specifically references DOS1, the other questions are not accessible as the framework is now closed, and was advised in a previous change not to update them because of such.

TheDoubleK:
Yeah changing `/questions/briefs` is legit, but changing `/questions/services` isn't.

At some point before DOS2 goes live for buying we will need to copy all DOS briefs changes across to DOS2 (assuming we don't want to regress when the framework surges over). But maybe better to do it in one fell swoop when the time comes rather than piecemeal.

danielknell:
Making the changes in both wont hurt though, would just speed up a wholesale porting of changes later?

TheDoubleK:
Either way, copy-and-paste a folder is quick enough :)
I'm not sure if other briefs questions have changed in just DOS too already (and there is the two types of brief that we won't need for DOS2). Fine if you want to make the corresponding changes now though.

danielknell:
there were changes in DOS2 i had to backport to DOS1 though, so wont quite be copy and paste, this solves the issue i was trying to fix, is it good for merge?

pcraig3:
Looks good to me 👍

PR #342, opened 2016-12-09
Don't include `type` in JSON schema for boolean with required value

idavidmcdonald:
Previously for a boolean question with a required value the JSON schema generated would require the value to match two conditions
- be a boolean
- match the required value

As the required value is already a boolean we don't need to have both of these conditions and can just have the later one (i.e. remove `type` from the JSON schema).

The reason for removing this is that we encountered a bug where if a value was a string then it would fail both of the above conditions. In python 3, due to randomised ordering of hashes we would then get a random error from the API (either picking the first or second condition). We could have got around this using logic in our API validation but it feels like this is a better solution as this should hold for all boolean questions with required values.


allait:
👍

PR #345, opened 2016-12-15
Edit brief response manifest to only have one section

idavidmcdonald:
This is because for the new brief response flow we will show one question per page rather than one section per page. All our questions will exist under one section rather than a section for each question.

Note, the section name will not be exposed so I have just chosen something reasonably appropriate.


Wynndow:
👍

PR #346, opened 2016-12-16
Make essential requirements met copy more consistent

idavidmcdonald:
We change "requirements" to "skills and experience" for consistency.

image


Wynndow:
How come we're only updating this small bit of content? As we need to do the rest of the content for the user research participants, shouldn't it all go in together?

idavidmcdonald:
I'm trying to put just these changes in and then it will mean that this pivotal story will be complete - https://www.pivotaltracker.com/n/projects/1884593/stories/129839965

There is a separate story for user research participants so hopefully this will enable us to get at least one story through the board that's been sat around for a while.

Wynndow:
cool 👍

PR #349, opened 2016-12-23
[#134368805]g7, g8 edit_service_as_admin: expand to cover all service fields that a supplier can set

risicle:
Story https://www.pivotaltracker.com/story/show/134368805

Though I talked briefly about destroying this file in favour of a symlink to `edit_submission.yml` in the end I didn't do this as I realized it needed some customizations over that.

Now it's *mostly* a direct copy of `edit_submission.yml` except for the inclusion of the `id` field and the merging of some sections in an attempt to reduce the number of single-question sections, while retaining the ordering of `edit_submission.yml`.


README.md
```diff
@@ -128,6 +128,8 @@ Then, whenever the framework content is changed:
 - alternatively, run `npm run frontend-build:watch` to automatically rebuild the framework content whenever a framework YML file changes
 - generate schemas into the API's `json_schema` directory using the `scripts/generate-schemas.py` script in this repo.
 
+Alternatively, if you think `bower link` is more trouble than it's worth, you can just symlink the your project's `bower_components/digitalmarketplace-frameworks` to wherever this repo lives on your local filesystem and it's 99% as good as the above.
```
allait: What's the problem with `bower link`? I'd rather we document one common way so that we don't have to go through a checklist of "how did you set this up" when someone has an issue with their local environment. And it seems to me that `bower link` is a more robust approach, since there's no way to know what bower might otherwise do to the contents of that folder on update.

risicle: It requires global (home-directory-wide) configuration in a nasty way and its complexity hides the fact that it's not *really* any more robust than `ln -s`. `ln -s` doesn't hide what it does behind a wall of 3 or 4 bad design decisions.

allait: `bower link` lets bower know what you're doing though, so it knows not to replace anything related to the dependency when you're running `bower update` after the version change. I'm pretty sure I had bower or one of our build steps rewrite my working directory when I was using similar manual approaches, so I'd be reluctant to suggest it unless you're confident it's safe.

risicle: i *think* bower just deletes the "directory" and hence the link when it "overwrites" it but I'd have to check.

risicle:
OK I've added a note of caution about symlinks in the readme now.

allait:
@risicle I'd argue that this addition is unhelpful at this point. You're advising people to do something that could potentially wipe out their working directory changes if they're not extra careful and the only argument for adding this is that you personally disagree with bower design decisions.

risicle:
I think your objection is absurd but I'm quite desperate to get this change in to limit the amount of rebasing I need to do of other PRs.

allait:
@risicle do you mind sharing why you think it's absurd?

risicle:
It's a comment which lets someone know something that took me a good amount of wading through google to find out, can save them a lot of time in setting things up in their home directory *just the way* bower wants it and gives them a good warning of what it could potentially do (which I'm not sure it does anyway). Developers are adults after all.

It's not important. Change ~~pooped~~ poopped out.

risicle:
Yay and now it's too late and I've got to rebase 3 pull requests.

pcraig3:
👍  

good luck with your rebasé.

risicle:
Muchos gracias.

PR #355, opened 2017-01-18
[#134368805]Revert "Revert "g-cloud-7, g-cloud-8 edit_service_as_admin...

risicle:
Story: https://www.pivotaltracker.com/story/show/134368805

Ok we should be safe to go with this now, accompanying admin frontend PR on its way too.


TheDoubleK:
:+1:

PR #356, opened 2017-01-19
Bug fix - example dates to be in the future

idavidmcdonald:
We don't want example dates to be in the past as per [this pivotal story](https://www.pivotaltracker.com/story/show/137360327)


Wynndow:
👍 after tests go green

PR #360, opened 2017-01-23
Remove full stop from availability

Wynndow:
For this story on Pivotal: [https://www.pivotaltracker.com/story/show/135599725](https://www.pivotaltracker.com/story/show/135599725)

If a buyer provides availability that ends with a full stop, you get two full stops. And it looks weird. See Pivotal story for an example.

Confirmed with Tracy that it's fine to remove full-stop completely.


Wynndow:
The horror.....

PR #361, opened 2017-01-25
fix ods questions

danielknell:
With all the rebasing etc of the  last branch it appears the  manifest with the  questions was not updated properly, it conntained all legacy questions, this fixes.


idavidmcdonald:
👍

PR #362, opened 2017-01-27
Add skeleton content for G-Cloud 9

TheDoubleK:
The dates are taken from the provisional G-Cloud 9 timetable detailed here: https://digitalmarketplace.blog.gov.uk/2017/01/19/g-cloud-9-a-provisional-timetable/

The declaration questions are taken from Digital Outcomes and Specialists 2.

There are a few basic service questions that will allow us to submit very basic "services" on each of the three G-Cloud 9 lots.


frameworks/g-cloud-9/questions/declaration/fraudAndTheft.yml
```diff
@@ -0,0 +1,14 @@
+question: >
+  Please confirm whether, within the past 5 years, anyone who represents, supervises or has control in your organisation or a partner or parent organisation has been convicted of any offence that relates to fraud affecting the European Communities’ financial interests as defined by article 1 of the Convention on the Protection of the Financial Interests of the European Communities, including:
+
+    - the common law offence of cheating the Revenue (HMRC)
```
benvand: Feels like `Cheating the (Public) Revenue` and `Conspiracy to Defraud` should be capitalised?

frameworks/g-cloud-9/questions/declaration/dunsNumber.yml
```diff
@@ -0,0 +1,10 @@
+question: What is your organisation’s DUNS number?
+type: text
+hint: >
+  The trading organisation applying to sell services through G-Cloud&nbsp;9 must have its own 9-digit DUNS number.
+  [Find out how to get a DUNS number.](http://www.dnb.co.uk/dandb-duns-number)
+
+validations:
+  -
```
benvand: It's a bit pedantic but we've got a mix of:
```
validations:
  -
    name: 'foo'
    message: 'bar'
```
and
```
validations:
  - name: 'foo'
    message: 'bar'
```

Don't know if there is a difference? Not really a dealbreaker.

frameworks/g-cloud-9/questions/services/allTheThingsToSee.yml
```diff
@@ -0,0 +1,26 @@
...
+  - label: Option 2
+    description: >
+       A description of option 2.
+    value: option_two
+validations:
+   -
+     name: answer_required
+     message: You messed it up you plonker.
+depends:
+  - "on": "lot"
```
benvand: The YAML writer has treated `on` as a reserved word here I think. AFAIK it's not.

benvand:
LGTM 👍  couple of comments but no real dealbreakers. Merge when ready.

frameworks/g-cloud-9/questions/services/allTheThingsToSee.yml
TheDoubleK: Thinking back I'm relatively sure that it's quoted because things broke when it wasn't.I'm 99% sure that these got quoted because we did see issues with it being treated as "true" at some point.  I can't remember the details though :/

frameworks/g-cloud-9/questions/declaration/fraudAndTheft.yml
TheDoubleK: The declaration content here is copied direct from DOS2 and is due for a complete review, so I'm tempted to leave this as it is for now.

frameworks/g-cloud-9/questions/declaration/dunsNumber.yml
TheDoubleK: There isn't any difference.  Would be nice to be consistent, I'll have a trawl through now.

frameworks/g-cloud-9/questions/services/lotName.yml
```diff
@@ -0,0 +1,2 @@
+question: 'Product type'
+type: text
```
galund: What is this question?

TheDoubleK: I'm 99% sure I know... all previous G-Cloud frameworks have it and I'm pretty sure it is a non-editable placeholder that is used to display the name of the lot that a service belongs to on the page.  e.g. pages for suppliers or admin editing a service, and maybe even the G-0Cloud service pages.  The service data gets updated with a `lotName` field, and then the manifest used to display the service can refer to the `lotName` "question" to show it on the page.

TheDoubleK: It may also be used for indexing G-Cloud services into Elasticsearch, so that lot names can be shown in search result summaries, eg.:
screen shot 2017-01-30 at 13 26 16

TheDoubleK: In short... I'm not sure exactly what would break if we don't have it, but I'm pretty sure *something* will.

PR #363, opened 2017-01-30
Split display_brief_response manifest into current and legacy versions

tombye:
From this point the page that show the results of an application to a brief needs to support applications with the old and new flow.

Splitting the manifest ensures both formats can be combined with the correct frameworks data.


idavidmcdonald:
👍

PR #368, opened 2017-02-03
Reorder output in ods file

danielknell:
need to move the  email field up the list of fields in the ODS.


idavidmcdonald:
👍 good when tests

PR #370, opened 2017-02-06
Update dos2 brief response questions and manifests to use new supplier flow

idavidmcdonald:
https://www.pivotaltracker.com/story/show/137518153

As we had been creating the 'new supplier flow' we only made
the changes to the dos1 content. The dos2 content will need to
reflect the new flow also.

Note, I have not copied across the legacy brief response questions
or manifests. This however relies on dos2 not being live until at
least 2 weeks after the new supplier flow gets released. If this
were not the case then we would also need to have the legacy
manifests and questions in the dos2 content.

Further note, if there are any more changes to dos1 brief response
content then we will need to update the dos2 content also to
reflect.


frameworks/digital-outcomes-and-specialists-2/manifests/edit_brief_response.yml
```diff
@@ -1,11 +1,10 @@
 -
-  name: Apply for this opportunity
+  name: Apply for an opportunity
```
Wynndow: Why the change to 'an'? On the buyer frontend `opportunities/brief_id` page we use 'this' still?

idavidmcdonald: Was a change carried over from when we made this new supplier flow manifest - https://github.com/alphagov/digitalmarketplace-frameworks/commit/47e4b545d75a4df23920f5af291a51b0c62b9df2


We don't actually show the name of the section anywhere for the new supplier flow so it shouldn't matter what the name is but maybe I'll change them both to be 'this' for consistency.

PR #378, opened 2017-02-15
schemas: allow "slug" attributes for manual slug specification

risicle:
(support for this already exists in content loader)

Please review my regex - the basic intention is "allow alphanumeric chars and dashes, not beginning or ending with a dash, but including the possibility of a single char slug".


risicle:
Ah yes the readme

PR #380, opened 2017-02-15
Bugfix spacing in declaration header

TheDoubleK:
Will fix this bug:
https://www.pivotaltracker.com/story/show/139811233


frameworks/g-cloud-9/manifests/declaration.yml
```diff
@@ -137,8 +140,9 @@
   slug: application-accuracy
   editable: True
   prefill: False
-  description: >
-    You must confirm the accuracy of the information in this declaration and in your service descriptions. We might ask for proof after you’ve submitted your application.
+  description: |
+    You must confirm the accuracy of the information in this declaration and in your service descriptions. We might 
+    ask for proof after you’ve submitted your application.
```
samuelhwilliams: Do we want to force a newline in here? It should either be > with newlines or | on a single line, I think.

frameworks/g-cloud-9/manifests/declaration.yml
```diff
@@ -165,8 +170,9 @@
   slug: organisation-registration
   editable: True
   prefill: True
-  description: >
-    The Crown Commercial Service will use the information you provide on this page to process your G-Cloud&nbsp;9 application. If your application is successful, it will carry out checks on your company.
+  description: |
+    The Crown Commercial Service will use the information you provide on this page to process your G-Cloud&nbsp;9 application.
+    If your application is successful, it will carry out checks on your company.
```
samuelhwilliams: As above.

frameworks/g-cloud-9/manifests/declaration.yml
TheDoubleK: How do you mean?  

It should be `description: | You must confirm...` rather than dropping it down?

PR #381, opened 2017-02-16
Follow up questions

allait:
### Split multiquestion schema generation

Adds helper methods to generate schema for flat (storing all properties
in a top-level object) and nested (storing properties in a nested object)
multiquestions.

Adds support for 'radios' question type to the followup schema
generation, and refactors it to use a `merge_schemas` helper to
make it easier to join multiple follow-up schemas.

Schema generation was using an older version of content-loader, which
didn't have DynamicList question class. Since DynamicList constructor
switches question type to 'multiquestion' when updating content-loader
we need to add additional dispatch logic to `multiquestion` so it can
call `dynamic_list` generator for `DynamicList` questions.

### Don't add lead question to the oneOf subschema required

Always listing lead-in question as required in a subschema makes the
validation fail when used with `enforce_required=False` in the API.

This wasn't a problem for brief responses, since both the lead-in and
the followup questions are nested inside one key, which is missing in
unrelated updates. But when storing them as separate top-level key we
need to be able to unset the required fields using the schema required
list.

So instead of listing the lead-in question as required in the subschemas
we add it to the main schema required list.

### Let multiquestions extend main schema fields

Allows property schema generators to return a schema addition to be
merged with the main schema. This enables multiquestions to add
allOf/oneOff subschemas for followup questions and modify the
required list on the main schema.

Changes the existing specialists anyOf/dependency additions to use
merge_schemas.

### Splits dynamic_list from nestd_multiquestion output

Dynamic lists are serialized as an array, but other multiquestions
aren't, so it makes sense to move the array part of the schema to
the `_dynamic_list` function.

And since we can no longer dispatch to the dynamic_list property
based on the question type we need to remove it from the types
mapping.


README.md
```diff
@@ -71,8 +71,7 @@ Question keys
 * `dynamic_field` defines the field in the filter context that will be used to generate dynamic list questions (eg when passing
   a brief as filter context setting `dynamic_field` to `brief.niceToHaveRequirements` will generate questions for nice-to-have
   requirements) (only valid for `dynamic_list` questions)
-* `followup` sets the name of the question that will be required/shown if the current question is answered with "Yes" (only valid
-  for `boolean` questions)
+* `followup` is a dictionary where each key is the name of the question that will be required/shown and the value for the key is a list of question answers that should display the follow up (only valid for `boolean` and `radios` questions)
```
samuelhwilliams: Followups are required for checkboxes, too.

schema_generator/utils.py
```diff
@@ -0,0 +1,16 @@
+def merge_schemas(a, b):
+    if not (isinstance(a, dict) and isinstance(b, dict)):
+        raise ValueError("Error merging '{}' and '{}'".format(
```
samuelhwilliams: This should probably be a TypeError, and the error message could probably be more descriptive: "Unable to merge disparate types: '{}' and '{}'."

allait:
@samuelhwilliams added support for checkbox question with follow-ups and optional follow-up questions. The tests need a new content-loader version to get merged, which I'm adding the checkbox support to now.

samuelhwilliams:
Probably need to add some tests around the followup schema generation.

allait:
@samuelhwilliams fair point, I've added a few tests

samuelhwilliams:
Other than that, from local testing and reading through, looks good. 👍

PR #385, opened 2017-02-17
G9 supplier declaration

cathrooney:
Adding additional G-Cloud 9 questions
Adding hint text about of environment, social and labour law
Hiding trade association questions if the supplier is established in the UK.


TheDoubleK:
@cathrooney I noticed the new `servicesHaveOrSupport.yml` was all G-Cloud 8, with four lots and everything.  I've had a quick stab at bringing it up to date for G9, but take a look before we merge this as it probably needs more tweaks (and I wasn't sure if the lots are numbered at all this time round?)

PR #388, opened 2017-02-24
140127375-supplier-declaration-content-tweaks

benvand:
Update G9 dates to be non breaking
Check BST vs GMT

https://www.pivotaltracker.com/n/projects/904880


galund:
Feels like this should be text processing done elsewhere (preferably in a template?) rather than changing the original data, as it's purely a presentational thing. (I mean, it could even be CSS, but if nbsp is easier that's fine.)

benvand:
As per chat with @allait nbsp here but we also need to add the end date as an attribute on model Framework.

TheDoubleK:
Oh this is pretty horrible.  We never had this issue with earlier frameworks... why now?  I can see one between "5pm" and "BST" might be desirable, but not the whole thing.

TheDoubleK:
I think we should maybe put a normal space between the time and date - after the comma - where there are both.

Otherwise, whatever :+1:

PR #390, opened 2017-02-27
Add names to a few declaration questions

samuelhwilliams:
Add names to the questions for the summary page and to support testing.


TheDoubleK:
These should be displayed on the declaration overview page in the summary table I think (pretty sure 'name' is used rather than 'question'?)

TheDoubleK:
At least, I think 'name' will be used in preference over 'question' if it exists.

samuelhwilliams:
Of course... :) I will chat to @andymcaleer

superroz:
**servicesDoNotInclude.yml**
1. It would be good to change the name from: 

'Other characteristics of your service'

to: 

'Services in scope for G-Cloud'

2. Add a ':' after include

3. Change: 'you are' to 'you're'.

**servicesHaveOrSupport.yml**
1. Change: 'you are' to 'you're'.

TheDoubleK:
I have some reservations about the names but :+1: to keep things moving.

superroz:
Cool. Cath and I need to go through the whole journey in the app so maybe
we can fix then. Would be good to either run through it with you or have a
chat to make sure we know what your main worries are.

PR #391, opened 2017-02-27
Markdown to html links

samuelhwilliams:
* Changes markdown links to HTML <a> tags and sets them to open in a new window.
* Bump version to 6.3.1


galund:
Don't really like this - assumes that in future people writing content will consistently remember to set link targets. Feels like this should be in the JINJA, preferably.

samuelhwilliams:
I don't necessarily disagree, but if we are going to have these links open in a new window for G-Cloud 9, this is probably the only way we're going to get it done - unless you know it's equally trivial to pass this off to Jinja?

PR #395, opened 2017-03-02
Tweak content on one of the declaration questions

samuelhwilliams:
* Tweak content and add sublist
* Bump version to 6.4.2


TheDoubleK:
Nested lists FTW :+1:

frameworks/g-cloud-9/questions/services/lot.yml
```diff
@@ -5,12 +5,21 @@ options:
   - label: Cloud hosting
     value: cloud-hosting
     description: >
-      Used to be IaaS and PaaS.
+      Cloud hosting services are cloud platform or infrastructure services that can help buyers:
+
```
TheDoubleK: Don't think you need an empty line before bullets, but will work OK like this too.

frameworks/g-cloud-9/questions/services/lot.yml
```diff
@@ -5,12 +5,21 @@ options:
...
+        - deploy, manage and run software
+        - provision and use processing, storage or networking resources
+
+      On G-Cloud 8, cloud hosting was known as ‘Platform as a Service’ and ‘Infrastructure as a Service’.
```
TheDoubleK: And G-Cloud 4, 5, 6 and 7!  But whatever.

TheDoubleK:
I probably wouldn't have bothered mentioning what the lots "used to be" - I thought the point of the new lots is that they are actually new, not just a new name for the same old things?  I only had "used to be SCS" etc. in before as a placeholder - I didn't think we'd actually say that.

Anyway, code is fine to merge if we're going with these words :+1:

samuelhwilliams:
Have just tried these lot descriptions but they don't currently work, as options.description isn't (and can't easily be) a TemplateField.

Have gone back to Laura for updated content; this is now on hold until that comes through.

pcraig3:
Last commit message isn't very descriptive :/

TheDoubleK:
Assuming these are the ones you've got working now :+1: again, for luck :shamrock:

PR #408, opened 2017-03-16
Message file about how submitted services are used

samuelhwilliams:
## Summary
Ticket: https://trello.com/c/RhoLFz3M/322-clarification-around-what-we-do-with-submitted-service-data

This PR adds a new message file to frameworks. This file supports a key, 'services', that will be displayed on the framework dashboard and clarifies how we use information provided by suppliers about their services (e.g. indexing it for search and making it public).

Also bumps version to 6.4.9

## Example
screen shot 2017-03-16 at 12 00 13


galund:
Can we have a link to the ticket in the commit message (or at least in a PR comment)? ta

TheDoubleK:
`use-of-data` feels maybe too specific as a top-level messages key - it can only really ever have one message in it?  Other messages files feel more generic... `urls`, `dates`, `homepage-sidebar`.

Not completely sure what would be better, but `use-of-data` feels more like it should be a key in the file rather than the top-level filename.

samuelhwilliams:
I spent some time trying to think of a better top-level name, but couldn't come up with anything. Open to suggestions.

I don't necessarily think it's too specific - it might be we want to advertise how we use different sections of data provided (e.g. declaration, services, supplier details, contact emails, etc).

TheDoubleK:
Given that I can't think of anything obviously better either, let's go with it.

:+1:

samuelhwilliams:
George/I just pondered over this and have made some changes that we think work better.

TheDoubleK:
Nice.

PR #417, opened 2017-04-21
Elasticsearch mappings and transformation generation

galund:
Script to generate Elasticsearch mapping for framework, inc transformations.

 - Most of the mapping comes, unchanged, from a template file which
   has been copied from the Search API i.e. mappings/services.json .
 - The main addition is of a 'transformations' node, which the
   Search API will use to preprocess the incoming service data before
   passing it on to Elasticsearch.
 - The only transformation we add so far is one that indexes parent
   categories relating to select subcategories. This is denormalization
   is done purely in the index.
 - Other meta information is added so that it's clear that the file
   in the Search API is no longer the original, and should not be
   edited by hand.
 - Brings in new version of Content Loader, which now depends on dm-
   utils library.

https://trello.com/c/etYh34Zs/341-fixing-software-parent-categories

### DIFF
--- a/schema_generator/search.py
+++ b/schema_generator/search.py
@@ -1,0 +1,78 @@
+from __future__ import print_function
+
+import os.path
+import sys
+import json
+from collections import defaultdict, OrderedDict
+
+from dmcontent import ContentLoader, utils
+
+
+_base_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+def _get_questions_by_type(framework_slug, question_types):
+    loader = ContentLoader(_base_dir)
+    loader.load_manifest(
+        framework_slug,
+        'services',
+        'search_filters'
+    )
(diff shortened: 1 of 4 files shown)


schema_generator/search.py
```diff
@@ -0,0 +1,78 @@
...
+    return [
+        OrderedDict((
+            ('field', checkbox_tree_question.id),
+            ('any_of', child_values),
+            ('append_value', sorted(ancestor_values)),
+        )) for ancestor_values, child_values in leaf_values_by_ancestor_set.items()
+        ]
+
+
+transformation_generators = {
```
samuelhwilliams: Can we uppercase this to make it stand out a little more as a global var?

galund: Sure, or should I give it a name starting with underscore? Probably not both I guess :)

PR #418, opened 2017-04-25
Use new date filter

benvand:


idavidmcdonald:
👍 good after tests green

idavidmcdonald:
👍 good after tests green

idavidmcdonald:
👍

PR #420, opened 2017-04-28
Update changelog for DOS2 breaking changes.

samuelhwilliams:
A new custom datetime filter has been applied to some DOS2 fields. The frontend apps fall over on init without manually updating the requirements.txt files for these repos to depend on newer util/content-loader versions.


TheDoubleK:
:+1: nice one.

PR #422, opened 2017-05-02
Fix description heading for summary page

galund:
Did this because of ticket http://trello.com/whatever


galund:
This was a test

PR #423, opened 2017-05-02
Update G9 framework agreement URLs

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/144692975

There's a chance that the "pdf" URL will change when a new version is uploaded, but there are right as of now.  Will wait for 100% confirmation before merging this.


galund:
It does seem super-fragile - I assume gov.uk has no way of automatically linking to the PDF version somehow? (I'd love a nice `rel=alternative` perhaps?)

When will the main URL be valid from - it isn't there yet?

TheDoubleK:
The page is switched on the same day as standstill starts.  But the URLs will be confirmed before then.  The fragility of the PDF URL is down to how GOV.UK publishes documents I think.

risicle:
I thought it was negotiated that we were going to be able to do the final for G9 in html...?

TheDoubleK:
There is an HTML version plus an accompanying PDF download I think, same as G8 and DOS2: 
https://www.gov.uk/government/publications/g-cloud-8-framework-agreement
https://www.gov.uk/government/publications/digital-outcomes-and-specialists-2-framework-agreement

Or was there a conversation that we wouldn't publish the PDF at all? I don't think I remember that?

PR #426, opened 2017-05-03
Search filter labels

galund:
* New fields 'filter_label' (for boolean q's) and 'filter_ignore' (for radio options)
* Also note that 'filter_label' option for radio/checkbox options will no longer be ignored by the buyer frontend.


frameworks/g-cloud-9/manifests/search_filters.yml
```diff
@@ -5,3 +5,18 @@
...
+-
+  name: User support
+  slug: user-support
+  questions:
+    - emailSupport
+    - phoneSupport
+    - webChatSupport
+-
+  name: Government security clearances
+  questions:
```
samuelhwilliams: If slug is needed for user support, it should probably be here for security clearances and security certification.

frameworks/g-cloud-9/questions/services/standardsCSASTAR.yml
```diff
@@ -2,7 +2,7 @@ name: CSA STAR certification
 question: >
   Do you have a current CSA Security, Trust & Assurance Registry (STAR) certification that covers the security of
   your service?
-
+filter_label: CSA STAR
```
samuelhwilliams: Is Roz and/or Karl OK with this acronym? We haven't presented it unexplained before and need to make sure we have their input.

frameworks/g-cloud-9/questions/services/webChatSupport.yml
```diff
@@ -22,10 +22,13 @@ type: radios
 options:
   - label: "Yes"
     value: "yes"
+    filter_label: "Web chat"
   - label: Yes, at an extra cost
     value: yes_extra_cost
+    filter_label: "Web chat (extra cost)"
```
samuelhwilliams: Not sure if it's distinctly useful enough to have (extra cost) as a filter label. For these, we should probably just filter on yes/no. I don't expect they will want to filter based on extra costs as this point, when they don't know what the base cost is.

galund: Yeah, this one will go away in favour of a new derived question, per <https://trello.com/c/BvnMwJZa/431-combining-web-chat-support-filter-options>. Just wanted to get something in there for demo purposes really.

frameworks/g-cloud-9/questions/services/standardsCSASTAR.yml
galund: It's in the current draft like this, I expect this all to change.

Acronyms are interesting, arguably if you want to filter on it, you know the acronym, but filters we think are particularly valuable, we clearly need to explain, somehow. Karl's thinking about it anyway.

frameworks/g-cloud-9/manifests/search_filters.yml
galund: Yup, not sure what it's for, actually. Will check...

galund:
Don't think I should merge this without adding the change to the search mapping that actually makes these new filters work.

frameworks/g-cloud-9/questions/services/emailOrTicketingSupport.yml
```diff
@@ -19,10 +19,13 @@ type: radios
 options:
   - label: "Yes"
     value: "yes"
+    filter_label: "Email/online ticketing"
```
samuelhwilliams: Generally we only use quotes around YAML words like yes/no/on/off which are otherwise interpreted as booleans. Longer strings aren't quoted in many other places.

PR #430, opened 2017-05-09
Support filters

karlchillmaid:
These are the three least complex filter labels for Cloud Support filters on G9 MVP

Ticket: https://trello.com/c/3XUKd0tX/345-add-filters-to-search-results-page

### DIFF
--- a/frameworks/g-cloud-9/questions/services/resellingType.yml
+++ b/frameworks/g-cloud-9/questions/services/resellingType.yml
@@ -17,12 +17,16 @@ type: radios
 options:
   - label: I’m not a reseller
     value: not_reseller
+    filter_label: Supplier is not a reseller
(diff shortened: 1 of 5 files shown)


frameworks/g-cloud-9/questions/services/resellingType.yml
```diff
@@ -17,12 +17,16 @@ type: radios
 options:
   - label: I’m not a reseller
     value: not_reseller
+    filter_label: Supplier is not a reseller
```
samuelhwilliams: We may want to store these without capitalisation on the first letter so that it looks more natural if we're going to describe searches as normal sentences (e.g. '1141 results found in Cloud hosting with staff security clearance that conforms to BS7858:2012').

PR #431, opened 2017-05-12
Support and software filters

galund:
Filters and indexing for G9 Cloud Software and Cloud Support lots.

https://trello.com/c/3XUKd0tX/345-add-filters-to-search-results-page


samuelhwilliams:
Can we squash these commits a little please... :)

frameworks/g-cloud-9/questions/services/freeVersionTrialOption.yml
```diff
@@ -1,6 +1,6 @@
 name: Trial option
 question: Do you provide a free trial option for your service?
-
+filter_label: free trial
```
samuelhwilliams: Can we keep a linebreak after the filter_label so they stay in four separate blocks, cheers... :pedant:

frameworks/g-cloud-9/questions/services/publicSectorNetworksTypes.yml
```diff
@@ -15,13 +15,19 @@ type: checkboxes
 options:
   - label: Public Services Network (PSN)
     value: psn
+    filter_label: Public Services Network (PSN)
```
samuelhwilliams: Guess it doesn't matter now, but we don't really need to add filter_labels if they are the same as the existing label. Would be nice to only override if it there's a difference.

karlchillmaid: I'm going to remove the ones like this where they are exactly the same. Am going to leave lower-case variants for now, though.

PR #432, opened 2017-05-17
Remove cloud-software lot from cloud-hosting questions

samuelhwilliams:
## Summary
A couple of questions in the repo accidentally got tagged as cloud-software questions, even though they should only appear for cloud-hosting. We do not believe this has had any outward effect on what buyers/suppliers have seen due to the ID being correctly overridden and the way the manifests are ingested, but it is still good to remove them to reduce any confusion in the future.

I've scanned the repo looking for other files with incorrect lots and didn't see any.

Bumps version to 8.3.2


samuelhwilliams:
Beaten to the punch by George: https://github.com/alphagov/digitalmarketplace-frameworks/pull/431

PR #436, opened 2017-05-17
Derive implied ordering for some filters

samuelhwilliams:
## Summary
Some of the questions we have asked list answers in order of preference, or where one answer inherently implies another answer (for example, if staff security are willing to undergo Developed Vetting, they are most certainly willing to undergo lesser clearances). We will therefore derive 'lesser' answers for search indexing, where suppliers have given the 'higher' answer.

This will ensure that all relevant services are returned if a buyer only ticks one of the checkbox filters (e.g. BPSS) and does not also click the 'higher' ones (e.g. DV, SC).

## Ticket
https://trello.com/c/zVcs1UON/432-search-filters-where-there-s-an-order-to-the-options


frameworks/g-cloud-9/questions/services/penetrationTesting.yml
```diff
@@ -18,10 +18,20 @@ options:
...
       value: at_least_once_a_year
+      derived_from:
+        question: penetrationTesting
+        any_of:
+          - at_least_every_6_months
     - label: Less than once a year
       value: less_than_once_a_year
+      derived_from:
+        question: penetrationTesting
+        any_of:
```
galund: I don't think this works without new filter labels, will mention to Karl.

galund: Actually maybe we just drop this one, says @karlchillmaid

PR #438, opened 2017-05-18
Remove invalid lot

samuelhwilliams:
## Summary
The `backup` question has cloud-software as a dependency, but this question is not asked of suppliers because it's excluded by the wrapping multiqBackup. We need to remove this lot so the filter doesn't appear for software searching.

Also fix one of the filters.

Bumps version to 8.4.4


TheDoubleK:
I checked 100 actual cloud software services and none of them have `backup` as a key, so this looks good.

:+1:

PR #441, opened 2017-05-19
Hosting filters

karlchillmaid:
Added and updated Hosting filters for G9 MVP

Ticket: https://trello.com/c/3XUKd0tX/345-add-filters-to-search-results-page


frameworks/g-cloud-9/questions/services/securityGovernanceStandards.yml
```diff
@@ -15,8 +15,10 @@ type: checkboxes
 options:
   - label: CSA CCM v3.0
     value: csa_ccm
+    filter_label: CSA CCM version 3.0
```
samuelhwilliams: This could probably just be changed in the label itself, so we don't need to add the filter_label key.

frameworks/g-cloud-9/questions/services/securityGovernanceStandards.yml
```diff
@@ -15,8 +15,10 @@ type: checkboxes
...
   - label: ISO/IEC 27001
     value: iso_iec_27001
+    filter_label: ISO/IEC 27001
```
samuelhwilliams: We don't need to add filter_labels if they are the same as the label itself.

frameworks/g-cloud-9/questions/services/scalingType.yml
```diff
@@ -12,8 +12,10 @@ type: checkboxes
 options:
   - label: Scales automatically
     value: automatic
+    filter_label: automatic
   - label: Scales with user intervention
     value: user_intervention
+    filter_label: manual through user interface
```
samuelhwilliams: Does this fully encapsulate the answer suppliers might have selected? Not sure if 'user intervention' might also be a phone call, which wouldn't quite make sense calling a 'user interface'?

frameworks/g-cloud-9/questions/services/securityGovernanceStandards.yml
karlchillmaid: When discussed with Roz the preference was for version to be written in full.

Should we keep the filter_label or edit the label (something we haven't been doing so far)

frameworks/g-cloud-9/questions/services/securityGovernanceStandards.yml
karlchillmaid: I'll remove this

frameworks/g-cloud-9/questions/services/scalingType.yml
karlchillmaid: Good point. I hadn't changed the wording on this one from the original draft, so assumed that 'through interface' would be ok. I'll have a think and reword it.

frameworks/g-cloud-9/questions/services/securityGovernanceStandards.yml
samuelhwilliams: Probably better to edit the label itself, if we're saying the content deserves to be changed. Otherwise we have two different versions of the same thing.

PR #445, opened 2017-05-22
Fixes for G-Cloud 9

samuelhwilliams:
## Summary
Minor tweaks for G-Cloud 9.
Bump version to 8.4.11

PR #447, opened 2017-05-22
Lot name indexed

galund:
We don't currently index top-level lot names as part of services, so if you search for "cloud hosting" keywords in the "Cloud Hosting" lot you lose about two thirds of the results.

https://trello.com/c/pTwwL1ht/466-index-lot-names-for-services


samuelhwilliams:
Cherry-picked into https://github.com/alphagov/digitalmarketplace-frameworks/pull/445

PR #448, opened 2017-05-22
Bump mock to 2.0.0

idavidmcdonald:


idavidmcdonald:
Just realised I haven't versioned this. I assume I don't need to because it's just a test dependancy I'm updating rather than anything that people would pull into apps?

TheDoubleK:
Sorry missed the comment - no need for a version bump, this never gets pulled in as a python package so the change here will have zero impact on apps that use it, only on tests for this repo.

PR #453, opened 2017-05-30
Remove duplicated key from service page

samuelhwilliams:
## Summary
Removes a duplicated key `userAuthenticationDescription` from the display_service manifest. Adds a schema validation to enforce unique items within a single group; better than nothing. Unfortunately I don't see a way to enforce uniqueness globally.

## Ticket
https://trello.com/c/WtXwWwl8/507-authentication-field-repeating-in-manifest


TheDoubleK:
:+1:

PR #457, opened 2017-06-13
Add DOS2 manifest for editing your service

idavidmcdonald:
https://trello.com/c/iptxy2Oy/397-allow-dos-suppliers-to-view-their-services

Note, the manifest has no editable questions and will only be used for viewing the service


frameworks/digital-outcomes-and-specialists-2/manifests/edit_service.yml
```diff
@@ -0,0 +1,125 @@
...
+
+- name: User research participants essentials
+  editable: False
+  questions:
+    - anonymousRecruitment
+    - manageIncentives
+
+- name: Team capabilities
+  editable: False
+  edit_questions: True
```
TheDoubleK: What does the `edit_questions` key do? Feels like this should probably be false too? Or is it to do with summary table layout, I can't remember exactly.

idavidmcdonald: From the frameworks repo README:
`editable` whether section allows updates for the questions, boolean value
`edit_questions` controls whether individual questions can be edited separately (only supported by multiquestion questions)

Turning on editing and keeping edit_questions `True` I wasn't able to see any differences in our pages.

I think `edit_questions` is redundant due to the frontend template (it specifies that the framework must be open to show edit links for individual questions) but my initial reasoning for keeping edit_questions was that if we do turn editing on, we only need to turn `editable` to `True` and then we will have all the existing editing behaviour (like when a supplier created their services) without forgetting that certain questions we allow to be edited individually. It looks like we would need to make changes to our template anyway if we wanted to have exactly the same functionality as when creating services.

I think I'm going to lean towards just removing the key.

benvand:
LGTM 👍

frameworks/digital-outcomes-and-specialists-2/manifests/edit_service.yml
samuelhwilliams: `edit_questions` is used by during the submission process to show links per-question rather than per-section (i.e. the one-question-per-page flow).

PR #458, opened 2017-07-07
Remove third party filter

karlchillmaid:
Removed third party user support filter.

Bumped version to 8.8.0

Ticket: https://trello.com/c/fVpk5C97/468-remove-user-support-available-to-third-parties-filter


VERSION.txt
```diff
@@ -1 +1 @@
-8.7.0
+8.7.1
```
galund: This is picky, but the minor version is the middle one, so we should have `8.8.0`.

The 'patch' version (the last one) is for bugfixes, this is kinda more than just a bugfix.

karlchillmaid: Not picky at all. I should've double checked the versioning. Will amend now.

karlchillmaid:
Thanks (and sorry!)

PR #460, opened 2017-07-14
Make features and benefits editable by suppliers

TheDoubleK:
Having `editable: False` means there's no edit link shown on the supplier editing service page.

Before:
screen shot 2017-07-14 at 14 51 14

After:
screen shot 2017-07-14 at 14 50 08


MuriloDalRi:
👍 :shipit:

PR #461, opened 2017-07-18
Gov security clearances split

galund:
Splitting the filter version of the question from the original means that we can preserve the original wording on the service page.


karlchillmaid:
Changes look as expected based on our conversation 👍

PR #462, opened 2017-07-27
Add stemming

samuelhwilliams:
## Summary
We want to be a little forgiving with user's search terms so that they don't miss relevant services if they're searching with a word in the wrong form (e.g. `hosted` instead of `hosting`). Therefore we'll apply a light stemming algorithm to search terms to collapse them to a root form (e.g. `hosting`/`hosted` will be reduced to `host` and indexed/searched against that term).

We will apply this against our main full-text service-specific fields: description, features, benefits, and as a bonus, categories (where it's easy to imagine users may search using a mutation of our term).

-We are also disabling sharding because our G-Cloud index is tiny (~60MB) and sharding is overkill. This will give us a little efficiency bump hopefully (albeit probably negligible).-

## Ticket
https://trello.com/c/Y6rdUgxU/517-search-stemming


Wynndow:
How does this get into the search-api? Search-api doesn't have frameworks as a dependency, and I can't seem to find a scripts which generates `/mappings/services.json`

frameworks/g-cloud-9/search_mapping.json
```diff
@@ -1,4 +1,27 @@
 {
+  "settings": {
+    "index": {
+      "number_of_shards": 1
```
Wynndow: Does dropping to one shard have any implications for data recovery if we lose a node? It makes sense that if we have the same 1 shard duplicated across three nodes it should be fine. I just wonder what the reasoning behind having 5 shards initially was?

samuelhwilliams: 5 shards is the Elasticsearch default, so I suspect it was just that.

We may need to increase replication so that each node gets a copy of the primary shard. I'm not sure how the number of replicas is assigned at the moment given we have different numbers of nodes in preview/staging vs prod. I will have to look into that.

samuelhwilliams: I'm going to take out the sharding change and stage that as a separate PR. Probably should have done that originally. So this PR will only be for stemming. :)

PR #469, opened 2017-09-14
Add become a supplier messages

TheDoubleK:
These are required for the new "become-a-supplier" page, see ticket: https://trello.com/c/kMYuzoge/75-create-the-become-a-supplier-page-2

The copies of the files for DOS, G6, G7 and G8 will never be get called in production, but it's handy to have them there for dev/preview/staging if people mess with framework statuses in those environments.

Highlights of the PR include use of the YAML reference, set with `&can_not_apply` and referenced using `*can_not_apply` to *Reduce Replication* within the message files.


risicle:
> Highlights of the PR include use of the YAML reference

Innnnteresting - are you sure we're not raising the barrier to editing by doing things like this?

schemas/messages.json
```diff
@@ -28,13 +28,22 @@
     "frameworkStates": {
       "properties": {
         "coming": {
-          "type": "object"
+          "type": ["object", "string"]
```
risicle: Does our code know what to do when handed an `object` in this field instead of a `string`?

TheDoubleK: It does in the places where it expects an object.

TheDoubleK:
> are you sure we're not raising the barrier to editing by doing things like this?

Certainly no more than by having embedded HTML lists.

schemas/messages.json
TheDoubleK: More specifically - these "frameworkStates" keys are used for a couple of different sets of messages: for the `homepage-sidebar` messages our frontend is expecting an object and knows what to do with it; the new `become-a-supplier` messages just have the straight-up string.
I think it's fine that the frontend will die if it gets the wrong thing, as I said in the commit message here: https://github.com/alphagov/digitalmarketplace-frameworks/pull/469/commits/bed9dc07c7416a387d24dced4d36a23c11022f55

risicle: Hmmmm okaaay, would seem it would be better to have it be a single-item dict in the simple case with the extra properties being optional but whatevs.

PR #473, opened 2017-09-23
Add search results download manifest

samuelhwilliams:
## Summary
We need to retrieve specific data fields from a service to insert into our search results download for Direct Award frameworks. This adds a new manifest which lists the question IDs in the order they should appear in the download file.

I've added the files for historical iterations because the manifest must be available for closed frameworks as well as live ones, and our current standard is to load manifests during start-up rather than on-demand, so the manifests must exist for all versions.

Bump version to 8.15.0.

## Required by
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/615

## Other associated PRs
https://github.com/alphagov/digitalmarketplace-api/pull/656
https://github.com/alphagov/digitalmarketplace-utils/pull/332
https://github.com/alphagov/digitalmarketplace-apiclient/pull/98
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/32

## Ticket
https://trello.com/c/gs2I3Cj2/676-end-your-search


TheDoubleK:
I think only G-Cloud 6 saved the `priceString` as part of the JSON.  For G7 and G8 we have :
"priceMax", "priceMin", "priceUnit" and "priceInterval" in the service JSON but the display string was then constructed from these elsewhere.

frameworks/g-cloud-9/manifests/download_results.yml
```diff
@@ -0,0 +1,5 @@
+- name: Body
```
TheDoubleK: Why does G9 get a capitalised `Body` but previous frameworks only get a small `body`?  Also, the name could maybe be more descriptive - "Body" doesn't really tell me what this is.

TheDoubleK:
Left some minor comments, see what you think.  The proof of the pudding is in the eating - if this is all working for you locally (multiquestions, etc.) then this is probably OK to merge.

frameworks/g-cloud-9/manifests/download_results.yml
samuelhwilliams: I could give it a more descriptive name, I suppose... but the name doesn't ever go anywhere.

I'll give it one anyway since you mentioned it!

samuelhwilliams:
Addressed comments. The content loader can manage it by iterating over sections and then getting all the fields for that section.

galund:
I don't think I follow why we need G6/7/8?

samuelhwilliams:
Because of how we/I load the manifests on the front-end (i.e. there's a loop that loads manifests for all G-Cloud iterations).

galund:
Got it. Yeah we should have some kind of concept of optional features or something, but not now.

PR #474, opened 2017-09-26
Flake8

benvand:
[flake8-report.txt](https://github.com/alphagov/digitalmarketplace-frameworks/files/1333609/flake8-report.txt)


.flake8
```diff
@@ -0,0 +1,12 @@
...
+# D203: 1 blank line required before class docstring
+# F401: 'identifier' imported but unused
+# W503: line break before binary operator
+exclude = venv*,__pycache__,node_modules,bower_components
+ignore = D203,W503
+max-complexity = 12
+max-line-length = 120
+putty-ignore =
+    **/__init__.py : +F401
+    scripts/ : +E402
```
samuelhwilliams: E402 isn't explained above

tests/test_generate_assessment_schema.py
```diff
@@ -12,56 +12,59 @@
...
-    "servicesDoNotInclude": True,
-    "servicesHaveOrSupport": True,
-    "termsAndConditions": True,
-    "termsOfParticipation": True,
-    "terrorism": False,
-    "understandHowToAskQuestions": True,
-    "understandTool": True,
-    "unfairCompetition": True,
-}
+def _definite_pass_g9_declaration():
```
TheDoubleK: What is the rule here?  "Don't use lambdas if you can avoid it"?

.flake8
TheDoubleK: Good catch.

tests/test_generate_assessment_schema.py
samuelhwilliams: The rule is something about not saving lambdas to variables.

benvand: https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
`E731: do not assign a lambda expression, use a def`

https://www.python.org/dev/peps/pep-0008/#programming-recommendations

It's to do with tracebacks being clearer.

PR #477, opened 2017-10-18
Restrict award contract value pricing fields to only take 0 or 2 decimal places

idavidmcdonald:
https://trello.com/c/PHnolEgP/16-change-money-format-on-awarding-a-brief-to-be-pounds-and-pence

Users had previously been able to say that a contract was worth £8.105. This should not be allowed. Instead we only allow positive numbers with 0 or 2 decimal places e.g £8 or £8.10. This new validation will be generated using the schema generator for the API.

The rest of our pricing fields remain the same and take up to 5dp.

Tests will fail until the content loader PR has been merged and requirements are updated in this PR.

Note for reviewer. I'm not sure if I'm happy with the key `decimal_place_restriction`. Let me know what you think. I'm also considering `pounds_and_pence` but open to any other ideas.


schema_generator/validation.py
```diff
@@ -271,8 +271,12 @@ def boolean_list_property(question):
     }}
 
 
-def price_string(optional):
-    pattern = r"^\d{1,15}(?:\.\d{1,5})?$"
+def price_string(optional, decimal_place_restriction=False):
+    pattern = r"^\d{1,15}(?:\.\d{1,5})?$"  # up to 5 decimal places allowed eg 90, 90.1, 90.12345
+    restricted_pattern = r"^[1-9](?:\d{1,15})?(?:\.\d{2})?$|^0\.(?!00)\d{2}$"  # 0dp or 2dp only allowed eg 90 or 90.12
```
samuelhwilliams: It would be nice if this restricted pattern were more forgiving about supplying decimals or not. It feels like we should still accept things like '90.' and '90.0', which still correlate to valid prices if we extrapolate appropriately. We want to enforce the new rules without making it less user-friendly.

Also we should keep the regexes semantically the same. We've added the `[1-9]` prefix at the start, so we should drop the repeat clause `{1,15}` down to `{1,14}` to keep us at the same overall max length.

idavidmcdonald: Good point on the `{1,14}`. Will implement.

idavidmcdonald: Comment from trello: After discussion with Sam and Tobi, we have decided to be more permissive with the input we are allowing.

If a user put 90. then we would allow this input and strip out the . before saving it. And if a user put 90.1, we would save it as 90.10.

The API validation will still remain as tight but the content loader will deal with the formatting before we send it.

tests/test_generate_validation_schema.py
```diff
@@ -347,6 +347,32 @@ def test_price_string():
...
+    invalid_prices = [
+        '0',
+        '0.00',
+        '150.1',
+        '150.000',
+        '150.00875',
+    ]
+    for price in invalid_prices:
+        m = re.search(price_string_validator, price)
+        assert m is None
```
samuelhwilliams: Can we inline these to get rid of the fairly pointless assignment to `m`?

PR #479, opened 2017-11-07
120 allow admin view and edit dos services update prices for digital specialists

benvand:
Single new manifest and version bump 
https://trello.com/c/U2Yd4CGD/120-allow-admin-view-and-edit-dos-services-update-prices-for-digital-specialists

Do not allow editing of service essentials


frameworks/digital-outcomes-and-specialists-2/manifests/edit_service_as_admin.yml
```diff
@@ -1,5 +1,124 @@
--
-  name: Description
-  editable: true
+- name: Lab name
```
TheDoubleK: ~~Might be worth grouping this with all the "Lab" questions towards the bottom of the manifest (only labs have `serviceName` defined).~~

Actually, this just follows the order in `edit_submission.yml` so let's stick with it as it is.

PR #481, opened 2017-11-14
add initial default.nix

risicle:
Much like all the other initial default.nix PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 __pycache__
 *.pyc
 .hypothesis/eval_source
-venv
+venv*
+local.nix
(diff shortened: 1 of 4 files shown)


galund:
I don't really like that all these have so much boilerplate / copy-pasta. Is there anything we can do to reduce that?

risicle:
Without creating yet another common repo and pulling in dependencies from it, not really.

galund:
Could they be in a repo that's already common across all the apps that this is being duplicated in. dm-utils maybe?

risicle:
It *could* but then we'd have to specify a certain *version* of that other repo, and that version wouldn't necessarily match the version that the main app depended on - we'd need to keep them in sync...

PR #483, opened 2017-11-17
Updated compare services URL

karlchillmaid:
Updated compare services URL.
Changes to headings in the G-Cloud buyers guide will break this link.

**This change shouldn't be released until the updated guidance is published w/c 20 November**

Ticket: https://trello.com/c/D6dmtXLC/177-fix-guidance-link-on-the-g-cloud-saved-search-task-list-page


galund:
Putting this live will require a change to the buyer app to pull in the new framework data, so make sure there's a dev task on the ticket and we know to pick that up.

galund:
(& therefore will need a version bump)

karlchillmaid:
Thanks @galund.

Shall I bump the version now, or wait?

galund:
If we might forget, make the change now. There's a chance someone else will also change the version number, and we'll get a conflict, though, so I don't mind... as long as we don't forget :)

PR #484, opened 2017-11-20
Brief search mappings

risicle:
Story https://trello.com/c/As7lMkmj/159-script-to-create-an-es-mapping-for-briefs

Here we make the changes required for generation of search mappings for entities other than `services` and also the ability for a framework to handle having search mapping for more than one search mapping at once.

This is done by moving `search_filters.yml` to `<entity name>_search_filters.yml`

As a result, an extra positional argument `entity_type` is now also required when calling `generate-search-config.py`

Also add initial mapping template for dos2.


risicle:
Looks like I might need to bump our travis python version (to 3.6?) to get the tests to pass for this.

risicle:
- [ ] Also looks like I need to update the readme to reflect these changes

risicle:
Cool - all references to `entity_name` should be replaced by `doc_type` now.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 Records breaking changes from major version bumps
 
+## 10.0.0
+
+The `search_filters.yml` manifest has been renamed to `<doc type>_search_filters.yml`. Code which referenced a
```
galund: This is a very backward not-compatible change. I guess it makes sense, but we'll need a buyer app PR and it would be reassuring to know you've checked this isn't going to be horrendous for some reason :)

schema_generator/search.py
```diff
@@ -77,30 +78,44 @@ def _derived_options_transformation_generator(checkbox_question):
...
-    mapping_json['mappings'][mapping_type]['_meta']['transformations'] = list(chain(
-        mapping_json['mappings'][mapping_type]['_meta'].pop("transformations", ()),
-        get_transformations(framework_slug),
+    original_meta = mapping_json['mappings'][mapping_type].get("_meta", {})
+
+    # starting our final _meta dict from scratch so we can ensure extra_meta gets the top spot, in ordered output
+    mapping_json['mappings'][mapping_type]['_meta'] = OrderedDict((
+        *((k, v) for k, v in extra_meta.items()),
+        # we want entries from original_meta to come *after* entries from extra_meta, but want to extra_meta entries
+        # to override original_meta, so ignore original_meta entries which are already in extra_meta
```
galund: This would be easier to read as procedural code mutter mutter

CHANGELOG.md
risicle: I think I remember you telling me "this is only ever referenced from the buyer app so it should be fine"...

galund: Yeah yeah, it's definitely only the buyer app. Hopefully there's only one place :)

schema_generator/search.py
risicle: Disagree *strongly*.

Procedural code would produce something that I was only 75% certain didn't do anything _funny_ without sitting down and mentally executing the code in my head. This describes what it wants and priority entries should have over each other.

scripts/generate-search-config.py
```diff
@@ -3,20 +3,19 @@
...
 
 To preview the mapping that will be generated, do not specify the output path.
 
-Note that the Search API only supports one mapping at a time, which will be
-used for any index relating to any G-Cloud framework. Therefore, care should
-be taken with the release process to ensure that indexing for the currently-
-live framework continues as expected, especially if the new framework's
-mapping is not backward-compatible with the old one.
+Note that most of the digital marketplace code only supports one mapping per
+doc type at a time. Therefore, care should be taken with the release process
```
galund: It might be more clear/correct to say "Note that most of the digital marketplace code only supports one index (and therefore one mapping) per doc type at a time."

And then add after this paragraph, perhaps:

>See <https://github.com/alphagov/digitalmarketplace-search-api/blob/master/README.md#updating-the-index-mapping> for more information about how to apply the updated mapping to an index. Backward-incompatible changes to the mapping should be applied by creating a new index, and swapping the index aliases over when ready (for example when deploying a frontend that references the new search manifest).

risicle: - [ ] do this

schema_generator/search.py
```diff
@@ -77,30 +78,44 @@ def _derived_options_transformation_generator(checkbox_question):
...
     print('', file=file_handle)
 
 
-def generate_config(framework_slug, extra_meta, output_dir=None):
-    mapping_type_name = 'services'
+def generate_config(framework_slug, doc_type, extra_meta, output_dir=None):
     if output_dir:
-        with open(os.path.join(output_dir, 'services.json'), 'w') as h_services_mapping:
-            generate_search_mapping(framework_slug, h_services_mapping, mapping_type_name, extra_meta)
+        with open(os.path.join(output_dir, '{}.json'.format(doc_type)), 'w') as base_mapping:
```
galund: Okay so we've agreed to include both `framework_slug` and `doc_type` in the file name, I think.

[ ] file name update :)

schema_generator/search.py
```diff
@@ -77,30 +78,44 @@ def _derived_options_transformation_generator(checkbox_question):
...
     print('', file=file_handle)
 
 
-def generate_config(framework_slug, extra_meta, output_dir=None):
-    mapping_type_name = 'services'
+def generate_config(framework_slug, doc_type, extra_meta, output_dir=None):
     if output_dir:
-        with open(os.path.join(output_dir, 'services.json'), 'w') as h_services_mapping:
-            generate_search_mapping(framework_slug, h_services_mapping, mapping_type_name, extra_meta)
+        with open(os.path.join(output_dir, '{}-{}.json'.format(doc_type, framework_slug)), 'w') as base_mapping:
```
galund: Oh, except we want them the other way around please ;)

frameworks/digital-outcomes-and-specialists-2/search_mappings/briefs.json
```diff
@@ -0,0 +1,152 @@
...
+  "settings": {
+    "analysis": {
+      "char_filter": {
+        "nonalpha_removal": {
+          "type": "pattern_replace",
+          "pattern": "\\W+",
+          "replacement": ""
+        }
+      },
+      "normalizer": {
```
Wynndow: So a normalizer is different to an analyzer, in that it guarantees that it produces a single token. But why is that important for us? Why can't we just use an analyzer?

scripts/generate-search-config.py
```diff
@@ -3,20 +3,25 @@
 
 At the time of writing, most of the search mapping is written by hand, in
 the search_mapping.json template file. This script adds to that template
-and writes it to a services.json file in the directory provided, which
+and writes it to a <doc_type>.json file in the directory provided, which
 should be reference the 'mappings' directory in a checkout of the Search API.
```
Wynndow: Can drop the word 'be' here.

schema_generator/search.py
```diff
@@ -77,30 +78,44 @@ def _derived_options_transformation_generator(checkbox_question):
...
+            generate_search_mapping(framework_slug, base_mapping, doc_type, extra_meta)
```
Wynndow: Should this be `generate_search_mapping(framework_slug, *doc_type*, base_mapping, doc_type, extra_meta)`

scripts/generate-search-config.py
risicle: shag.

schema_generator/search.py
risicle: well caught. yes.

schema_generator/search.py
```diff
@@ -77,30 +78,44 @@ def _derived_options_transformation_generator(checkbox_question):
...
-def generate_search_mapping(framework_slug, file_handle, mapping_type, extra_meta={}):
-    with open(os.path.join(_base_dir, 'frameworks', framework_slug, 'search_mapping.json'), 'r') as h_template:
+def generate_search_mapping(framework_slug, doc_type, file_handle, mapping_type, extra_meta={}):
+    with open(os.path.join(
+        _base_dir,
+        "frameworks",
+        framework_slug,
+        "search_mappings",
+        "{}.json".format(doc_type),
+    ), 'r') as h_template:
```
Wynndow: what does the `h` stand for?

galund: "handle". Probably naughty of me to sneak in some rotten old hungarian notation, but it's there now and I got away with it so :-p

schema_generator/search.py
```diff
@@ -77,30 +78,44 @@ def _derived_options_transformation_generator(checkbox_question):
...
         mapping_json = json.load(h_template, object_pairs_hook=OrderedDict)  # preserve template order for git history
 
-    mapping_json['mappings'][mapping_type]['_meta'].update(extra_meta)
-    mapping_json['mappings'][mapping_type]['_meta']['transformations'] = list(chain(
-        mapping_json['mappings'][mapping_type]['_meta'].pop("transformations", ()),
-        get_transformations(framework_slug),
+    original_meta = mapping_json['mappings'][mapping_type].get("_meta", {})
```
Wynndow: Does this need to be a copy so it's not mutated by what happens below?

schema_generator/search.py
risicle: do i spy an ex-windows programmer...?

schema_generator/search.py
risicle: ah (you're feeding these to me aren't you?) but "what's written below" is written in a functional style so we *know* (within about 99% probability) that it *doesn't* have any weird side effects like mutating any of its operands.

(the missing 1% is developers who are completely insane and sneak mutating code into it anyway but occurrences of that should stand out like a sore thumb)

schema_generator/search.py
galund: We all have a past, right... (to be fair I never did much C, so don't ask me about that)

schema_generator/search.py
galund: lol

Wynndow: I see. Nice.

frameworks/digital-outcomes-and-specialists-2/search_mappings/briefs.json
risicle: analyzers don't work with `keyword` field types

Issue #412, opened 2017-03-31
README.md linking for dev use instructions fail


benvand:
As far as I can make out this package shouldn't have a `node_modules` directory because it has no `node.js` dependencies.
https://github.com/alphagov/digitalmarketplace-frameworks/blame/master/README.md#L134

Therefore this won't work:
```
A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
as follows:

- from this repo, run `node_modules/bower/bin/bower link`
- from each app, run `node_modules/bower/bin/bower link digitalmarketplace-frameworks`
```

I've worked around this by installing bower globally and linking this package globally like so:
```
- from anywhere, run `sudo npm install -g bower`
- from this repo, run `bower link`
- from each app, run `bower link digitalmarketplace-frameworks` or `node_modules/bower/bin/bower link digitalmarketplace-frameworks`
```

I'm not much of an `npm` / `bower` guy so I'll need some input as to whether or not this has any caveats @galund

TheDoubleK:
Weird - I'm sure people have had the bower link stuff up and running in the past.

galund:
This is clearly my fault, I think I probably copy-pasted my new instructions from another repo (the instructions about this were missing from all of them).

You can certainly install bower globally, or alternatively use any of the bowers that you've installed elsewhere (if you give a full path). Feel free to update the readme, it'd be nice if it made sense. 

Otherwise, assign me this issue, and I might get around to it sometime :)

TheDoubleK:
I reckon it'd be nice to avoid recommending global bower.

benvand:
Maybe a  makefile command or something.
`make bower_link <my_local_frameworks>`

To be run from each of the frontends.

galund:
If `make` is to run bower, it still has to be able to find it somewhere :/

galund:
Oh, no I understood what you mean now. Yep.

carolinegreen:
I don't like installing things globally either but Bower is generally designed to be installed globally as it's a node command line tool.

https://docs.npmjs.com/getting-started/installing-npm-packages-globally

galund:
Saw this ticket and felt ashamed... I've updated the readme, I didn't include the `sudo` because it wasn't necessary for me, and hopefully wouldn't be for anyone else as long as npm is set up right.

https://github.com/alphagov/digitalmarketplace-frameworks/pull/485

galund:
Readme fixed up, merged.

PR #485, opened 2017-11-28
Suggest use of global bower instance.

galund:
https://github.com/alphagov/digitalmarketplace-frameworks/issues/412


README.md
```diff
@@ -163,8 +163,9 @@ Development
 A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
 as follows:
 
-- from this repo, run `node_modules/bower/bin/bower link`
-- from each app, run `node_modules/bower/bin/bower link digitalmarketplace-frameworks`
+- install bower globally (if you didn't already): `npm install -g bower`
+- from this repo, run `bower link`
+- from each app, run `bower link digitalmarketplace-frameworks`
```
TheDoubleK: Is it worth also adding instructions how to UNLINK once you're done.  In my experience having things linked locally causes more problems than it's worth, so knowing how to undo this damage is useful.

Also, what's the thinking behind global rather than local bower? (I have no idea about this at all)

Also (I know not part of this PR), is it worth removing the bit about "(NB this step also required before deployment to PaaS via `cf push`)"?  The Frameworks repo seems like a really weird place to have that information, and it's not something we should be encouraging anyone to do much.

MuriloDalRi: Apparently, Bower is supposed to be installed globally. See Caroline's comment: https://github.com/alphagov/digitalmarketplace-frameworks/issues/412#issuecomment-291151394

_Hopefully_ we'll get rid of bower quite soon!

galund: Specifically global rather than local because this repo doesn't have a local place to install bower (which is why the instructions were wrong in the first place).

I reckon people can probably look up the docs to unlink, I'm not sure adding something here would be enough of a reminder (i.e. that this could be the cause) when things go wrong.

PR #487, opened 2017-12-04
101 suppliers should be able to select a category es for frontend

Wynndow:
Part of this ticket: https://trello.com/c/KcqU7k20

### Add dmfilter_status to briefs search mapping
To be able to filter by status on the frontend, we need to have status
with the filter prefix in the search mapping.

### Add status to the `briefs_search_filters` manifest and question
This allows the frontend code to generate a filter for the list
opportunities page.


frameworks/digital-outcomes-and-specialists-2/manifests/briefs_search_filters.yml
```diff
@@ -1,10 +1,12 @@
...
+  questions:
+    - status
 -
   name: Location
   slug: location
   questions:
-    - locationOutcomes
-    - locationParticipants
-    - locationSpecialists
+    - locationAllLots
```
galund: If this became a regular thing, it might be nicer to have a way of doing this automatically, e.g. (and you'll notice I haven't thought this through at all) something like `ignore-depends-on-restriction: ["lot"]`.

But we certainly don't need to bother if the only example is now this one field.

Wynndow: Yeah, I agree. Have had a look through though and this is the only example of this I can find.

frameworks/digital-outcomes-and-specialists-2/manifests/briefs_search_filters.yml
```diff
@@ -1,10 +1,12 @@
+-
+  name: Status
```
galund: Don't think we should be adding this until this ticket has actually been kicked off: https://trello.com/c/LFZSi6r1/105-suppliers-should-be-able-to-filter-by-more-statuses-of-an-opportunity

In particular, the Acceptance Criteria on there talk about having a feature flag. Laura might get rid of that requirement

The thing I'm more worried about is whether people using this interface will properly understand the limitations of that data - this ticket never made it into any sprint :-( https://trello.com/c/ARcnsQLM/133-suppliers-need-to-understand-the-completeness-of-status-data

frameworks/digital-outcomes-and-specialists-2/manifests/briefs_search_filters.yml
TheDoubleK: Is that not just `location` if the question is identical for all lots?

TheDoubleK: The only reason for splitting questions by lot is if they have different wording.

Wynndow: Hey Kev, the locations for specialists and outcomes are slightly different to user research participants. They have 'offsite', and user research participants has 'international'. So we need the lot specific questions for when creating a brief.

It was decided that rather than having lot specific location filters, which would cause the available locations to filter by to change when moving between lots, that there should be one filter which covers all of the available options. Even if that means that you always get 0 results for user research participants if applying the offsite filter, for example.

This is the question that powers the filter, with all values and not depending on lot. Do you think there might be a better way of doing it?

You might be right about calling it `location`, although i was following the pattern of <question><lot>.

frameworks/digital-outcomes-and-specialists-2/manifests/briefs_search_filters.yml
Wynndow: We had a chat about this at standup and decided to stick with just open/closed for now, so I'm going to remove.

frameworks/digital-outcomes-and-specialists-2/questions/briefs/locationAllLots.yml
```diff
@@ -0,0 +1,20 @@
...
+  - label: Yorkshire and the Humber
+  - label: East Midlands
+  - label: West Midlands
+  - label: East of England
+  - label: Wales
+  - label: London
+  - label: South East England
+  - label: South West England
+  - label: Northern Ireland
+  - label: International (outside the UK)
```
galund: Totally struggling to imagine what this even means in the context of DOS, but eh

Wynndow: I think user research participants can be international. If that's ever actually happened, I don't know.

galund: Participants wanted in Hawaii. No particular reason... :-D

DilwoarH: It would be good to have values against the options as well, is this something that is possible?

The reason why I say this is because these labels have spaces and when we send virtual page views, we use this information to construct URLs (not very friendly when the URLs contain spaces)

galund: We have to go with whatever's stored in the API, so this really has to match "location.yml" file. I agree it's a pity we don't automatically construct a prettier value for this things, but that'd involve some quite complex changes across the content loader and the apps :(

galund: (Hmm actually maybe we strip spaces anyway, so maybe it's just a question of changing the frontend to remove the spaces in _its_ URLs. Either way, no change required here.)

Wynndow:
If you have a label without a value, as we do here, then the value is automagically whatever the label is. As George said, the label needs to match what's in the api, which is these locations with spaces. So the only thing we could do is have a different label with the values as listed above.

frameworks/digital-outcomes-and-specialists-2/questions/briefs/locationAllLots.yml
katstevens: Don't we slugify things like this when creating URLs?

Wynndow: I don't think we use these values in URL's anywhere, but they might be part of the URL's for the virtual page views. I'm not very familiar with them to be honest. @DilwoarH ?

galund: They would be in the URLs for when we search by location. But if we don't like the look of those, that's a buyer-app fix, not a change here, I think.

DilwoarH: yeah agreed with you @galund 

Thanks for addressing the comment.

PR #496, opened 2018-02-14
Remove bower link from readme

galund:
Remove reference to `bower link` which won't work any more.
 - ideally we'd replace this flow with `yarn link`, but that doesn't work - will raise a tech debt ticket.

For reference, `yarn link` wants the frameworks repo to be an NPM-comptaible package, which it probably could become:

```
$ yarn link
yarn link v1.3.2
error Couldn't find a package.json file in "dev/digitalmarketplace-frameworks"
info Visit https://yarnpkg.com/en/docs/cli/link for documentation about this command.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,18 +169,12 @@ ES multi-fields are for, but unfortunately those don't turn out to be quite flex
 Development
 -----------
 
-A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
-as follows:
+Whenever framework content is changed, you will need to do the following:
 
-- install bower globally (if you didn't already): `npm install -g bower`
-- from this repo, run `bower link`
-- from each app, run `bower link digitalmarketplace-frameworks`
-
-Then, whenever the framework content is changed:
-
-- on each frontend app, rebuild the app's `content` directory to include the new content by running `make frontend_build`
-- alternatively, run `npm run frontend-build:watch` to automatically rebuild the framework content whenever a framework YML file changes
-- generate schemas into the API's `json_schema` directory using the `scripts/generate-validation-schemas.py` script in this repo.
+- on each frontend app, temporarily point the `package.json` to your branch of this repository, and rebuild the app's
+  `content` directory to include the new content by running `make frontend_build`
+- generate schemas into the API's `json_schema` directory using the `scripts/generate-validation-schemas.py` script in
+  this repo.
 
 Running the tests
 -----------------


README.md
```diff
@@ -169,18 +169,12 @@ ES multi-fields are for, but unfortunately those don't turn out to be quite flex
...
-- install bower globally (if you didn't already): `npm install -g bower`
-- from this repo, run `bower link`
-- from each app, run `bower link digitalmarketplace-frameworks`
-
-Then, whenever the framework content is changed:
-
-- on each frontend app, rebuild the app's `content` directory to include the new content by running `make frontend_build`
-- alternatively, run `npm run frontend-build:watch` to automatically rebuild the framework content whenever a framework YML file changes
-- generate schemas into the API's `json_schema` directory using the `scripts/generate-validation-schemas.py` script in this repo.
+- on each frontend app, temporarily point the `package.json` to your branch of this repository, and rebuild the app's
```
idavidmcdonald: might be nice to be explicit that you must push your branch up

README.md
```diff
@@ -169,18 +169,12 @@ ES multi-fields are for, but unfortunately those don't turn out to be quite flex
 Development
 -----------
 
-A local checkout of the frameworks repo can be shared with locally-running services (i.e. frontend applications)
-as follows:
+Whenever framework content is changed, you will need to do the following:
```
idavidmcdonald: I find `Whenever framework content is changed` a little misleading. Maybe something like `If you want to use a development branch of the frameworks repository in one of the other applications`

README.md
idavidmcdonald: Also I'm not sure if `make frontend_build` is now `make frontend-build`.

Also I'm not sure if the frontend-build will also install that new branch? I haven't spent long enough looking to see if you also need to run some kind of yarn install command first...

TheDoubleK: I don't think it was ever `frontend_build`, though I've only recently moved to using the Makefile so could be wrong.  It's definitely hyphenated now though, and the NPM version I used was always `npm run frontend-build`

galund: Yeah, it changed last year :) https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/773f6e704a5ddf7d5d35c0bd55af43821369f93b

But yeah @idavidmcdonald you're right, `frontend-build` isn't enough to get the new version.

In fact `make yarn-install` doesn't work either, you have to update your lockfile first.

So actually you need to run `yarn install` followed by `make frontend-build`.

But I'd kinda prefer to document what people actually _do_ rather than something hypothetical. I haven't been editing framework data in a while: @TheDoubleK @katstevens any thoughts on what we should actually have here? I can just remove the entire section if it's not useful (I added it in the first place, to document what I had been shown).

katstevens: Adding my remote branch to `package.json`, then doing `yarn install && make frontend-build` works for me.

idavidmcdonald: I used to use the bower link behaviour and ideally would still continue to do so (via yarn) without having to commit, push up a branch, install the requirements every time I want to make a trivial change...

galund: Cool, okay that makes me think it might be worthwhile to get `yarn link` working rather than faffing around with the wording here too much more :-)

samuelhwilliams:
Bump.

galund:
Closing this as making `yarn link` work seems to have been super-trivial - https://github.com/alphagov/digitalmarketplace-frameworks/pull/508

PR #497, opened 2018-02-14
Dh gdpr contract update

DilwoarH:


TheDoubleK:
It seems strange to me that the acronym "LED" isn't defined until near the bottom.  Usually you define an acronym in full when it's first used?

DilwoarH:
@TheDoubleK https://github.com/alphagov/digitalmarketplace-frameworks/blob/ac57d69e60dded48719a3edcf2268263a2219956/frameworks/g-cloud-9/messages/contract_variation_1.yml#L192

frameworks/g-cloud-9/messages/contract_variation_1.yml
```diff
@@ -18,7 +18,7 @@ variation_comes_in_effect_description: We’ll tell you when CCS has countersign
 
 variation_agreement_label: I have the legal authority to accept these changes and agree that the company will be bound by them.
 
-confirmation_message: You have accepted the proposed changes. They will come into effect when CCS has countersigned them. We’ll let you know when this happens.
+confirmation_message: You've accepted the proposed changes. They'll come into effect when CCS has countersigned them. We’ll let you know when this happens.
```
samuelhwilliams: These should be fancy apostrophes (`’`). Copy-pasted from the `CCS may amend this [...]` bit below. We (try to) use these rather than bog standard apostrophes because content prefers the look of them.

DilwoarH:
@TheDoubleK @samuelhwilliams @idavidmcdonald - all comments addressed

TheDoubleK:
Yep I know it's defined - but on line 197, but it's already been used on lines 103 and 161.

PR #499, opened 2018-02-23
briefs text search

risicle:
This is a mapping skeleton to produce a mapping approximately equivalent to what we agreed to try in https://trello.com/c/qdNbRJ4e/331-supplier-can-search-for-buyer-on-dos

Most of the interesting parts are just cribbed straight from service/g-cloud search.


frameworks/digital-outcomes-and-specialists-3/search_mappings/briefs.json
```diff
@@ -137,20 +137,29 @@
...
+          "type": "text"
         },
         "dmfilter_specialistRole": {
           "type": "keyword",
           "normalizer": "filter_normalizer"
         },
         "dmtext_summary": {
-          "type": "keyword"
+          "type": "text",
+          "analyzer": "stemming_analyzer"
```
katstevens: In the G-Cloud mapping, `stemming_analyzer` is defined as an `analyzer`. Do we need to do it here too?

katstevens: Also looking at the G-Cloud mapping: is `"term_vector": "with_positions_offsets"` the magic sauce that does the search highlighting in the G Cloud `serviceDescription`?

risicle: Damn it I didn't include those sections in the commit for some reason.

risicle:
ok - this new version works better, it's now purely the buyer frontend's fault that we're not yet getting *visible* highlighting...

PR #501, opened 2018-03-05
Give users access to the contract variation as a document

DilwoarH:
Ticket: https://trello.com/c/zz0CwpNu/134-give-users-access-to-the-contract-variation-as-a-document


DilwoarH:
Need to change commit message to say bump

PR #505, opened 2018-03-09
Add more cyber security accreditations

TheDoubleK:
For this ticket: https://trello.com/c/Khm9AS0X/68-re-add-cyber-for-g-cloud-10

Suppliers with some specific security accreditations were barred from supplying through G-Cloud and instead directed to the "Cyber framework" instead.  But for G-Cloud 10 they will be allowed.

So we have been told to "re-add cyber" for G-Cloud 10.  

No-one really knows what this means, so I've had a go at figuring it out here.

Trawling the history of this repo I found two security-related questions that were deleted mid way during development of G-Cloud 9.  They look like this:
screen shot 2018-03-09 at 16 17 41

Looking at the way the security questions ended up for G-Cloud 9 launch it looks to me like these questions are unnecessary as full questions, and the same information can be collected from suppliers with the addition of two checkboxes to the list of accreditations that security services can have.

So I've done that, and removed the text telling suppliers with security accreditations that they can't apply to G-Cloud, because now they can.

## Existing security testing page - 🚫NO CYBER🚫 

screen shot 2018-03-09 at 17 22 18

## Proposed new security testing question page - :ok: CYBER :ok: 

screen shot 2018-03-09 at 17 10 00

### DIFF
--- a/frameworks/g-cloud-10/questions/services/securityTesting.yml
+++ b/frameworks/g-cloud-10/questions/services/securityTesting.yml
@@ -3,13 +3,6 @@ question: Do you provide security testing?
 question_advice: |
   Security testing services include penetration testing, IT Health Checks and risk analysis.
 
-  Don’t submit your services to G-Cloud if they’ve been assured by these National Cyber Security Centre (NCSC) schemes:
-
-    - Cyber Security Consultancy
-    - Penetration Testing (CHECK)
-    - Cyber Incident Response (CIR)
-    - Tailored Assurance Service
-
 depends:
   - "on": lot
     being:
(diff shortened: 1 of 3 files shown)


lauraflannery:
Hey @TheDoubleK and Patrick

I would check with Rob MacLeod about 2 questions:

'what kind of security testing do you provide'
- should we be giving more options?

'which security accreditations do you security testers have?'
- there are lots of schemes listed on the NCSC site (see here: https://www.ncsc.gov.uk/marketplace)
which ones should be listed on G-Cloud?

samuelhwilliams:
I believe we still need this to go in.

PR #506, opened 2018-03-12
Add transformation field for open closed briefs

brenetic:
This adds a new mapped field, `openClosedStatus` for the award status. It is a direct copy of what is already there as we need to use this to continue to provide the 'open' and 'closed' filtering for search results.

We also need to keep the original brief status in Elasticsearch so we can extract it to provide further detail to the end user in the search results.

The old field mapping can be removed once changes to the `buyer-frontend` have been made to use the new `statusOpenClosed` field.

Bump the version to 10.8.0

https://trello.com/c/y59qiKOl/335-show-award-status-in-dos-search-results


katstevens:
Before I delve any further, do we need to copy this over for DOS-3? And why don't we have the equivalent of `brief_search_filters` for DOS-1? I get that we only use the one DOS mapping in the Search API but I don't really understand why.

frameworks/digital-outcomes-and-specialists-2/search_mappings/briefs.json
```diff
@@ -62,6 +62,28 @@
...
+              "any_of": [
+                "awarded",
+                "unsuccessful",
+                "cancelled"
+              ],
+              "set_value": "closed"
+            }
+          },
           {
             "set_conditionally": {
```
katstevens: We're still mapping the statuses to closed, below? Should we remove that?

katstevens: (Ah - I see this is stage 1!)

VERSION.txt
```diff
@@ -1 +1 @@
-10.7.2
+11.0.0
```
katstevens: If this is step 1, does that mean that this isn't the breaking change?

frameworks/digital-outcomes-and-specialists-2/search_mappings/briefs.json
brenetic: These will be removed a little bit later. Doubling up momentarily will allow us not to break buyer-frontend's search filters

brenetic:
I am not sure about the answer to whether we should be copying this over to DOS-3 or why there is no equivalent filter for DOS-1. @galund or @risicle ?

risicle:
Yes it almost certainly should be copied to DOS3. Old ones like DOS are generally as they were when we stopped using them.

Technically the DOS2 mapping covers both DOS and DOS2 briefs and the DOS3 mapping will (almost certainly) be expected to support briefs from all three frameworks.

frameworks/digital-outcomes-and-specialists-2/questions/briefs/status.yml
```diff
@@ -1,9 +0,0 @@
-name: Status
-id: status
-question: status
-type: radios
-options:
-  -
-    label: open
-    value: live
```
katstevens: As discussed yesterday, we'll need to keep this in for the moment otherwise the Buyer FE won't render the filter form properly.

risicle: Is this intermediate version even intended to be pulled into the buyer FE?

katstevens: Maybe not by us, but someone else might! And it would certainly be a breaking change if we left this change in.

galund: I don't think this is a breaking change - if the buyer frontend is updated to point to the new frameworks version, then it will also be using the new field name - which we will have indexed.

galund: (It is breaking only insofaras the URL's will be changing, but we don't guarantee that as an API and I'm considering that to be acceptable, although contrary views welcome.)

risicle: *Massive shrug*

katstevens: After chatting I'm convinced it'll all be OK 😸

PR #508, opened 2018-03-19
`yarn link` fixed

galund:
Add package.json to make this repo yarn-linkable, and update the version numbering system to be easier to use (and work in an idiomatic way for a yarn repo).
 - yarn supported installing this package without it having a package.json, but it doesn't support linking, which is a useful thing (see README).
 - README updated to clarify how this now works, and improve the wording more generally;
 - use `yarn version` to update the version number.

Replaces https://github.com/alphagov/digitalmarketplace-frameworks/pull/496

https://trello.com/c/ua7JddmK/394-yarn-link-and-versioning-of-frameworks-repo


galund:
Incidentally (while I'm looking at the versioning of this repo), it looks like `tag_version.sh` has been slightly broken for a while, for example: https://ci.marketplace.team/view/Utils%20and%20toolkit/job/tag-frameworks/385/console

 - changelog it attempts to output relies on the output order from `git tag`
 - annoyingly, since we went from v9 to v10, we will have been comparing everything to v9, because v10 appears not at the bottom of the list but half way down (i.e. the order is lexographic rather than numeric)

Proposal to fix this is to get rid of that output, which I assume we aren't using anyway?

Am also going to suggest we use `yarn version` not to generate the tag, but to generate the commit message, as it seems to be capable of looking after this for us in a nice way (with some configuration).

galund:
(I've confirmed that `jq` is on there so I reckon this should work.)

TheDoubleK:
This seems like a nice way to do versioning, my only slight concern is that it's a new third different way to update version number on our libraries (`init.py` vs `version.txt` vs `package.json`).  I'm probably not the best person to ask whether this is a change that the team is happy to adjust to at this point.   But this does seem neater.

galund:
So my idea is that `version.txt` would just go away. Things that get installed with yarn go this way, things that get installed with pip go the `init.py` way. I don't think there's a third set of dependencies (is there?)...

PR #513, opened 2018-04-12
Change to certification noun

samuelhwilliams:
## Summary
Change language to consistently use 'certification' rather than
'accreditation' where appropriate (i.e. generally where 'accreditation'
is used as a noun or where we do not need to repeat the same/similar
word close together).


samuelhwilliams:
Hmm, hopefully this change is just 'plain English'ing, so my intent would be for them to be fine to still copy across. It's a good point, though...

constancecerf:
Yes, definitely not a material change! We can carry them over.

PR #515, opened 2018-04-12
Copying services metadata

samuelhwilliams:
## Summary
We want suppliers to be able to copy (published) services from one
framework (G-Cloud 9) to the next iteration (G-Cloud 10). In order to do
this we need to define, in the new framework, which source framework
contains compatible services. We then also need to whitelist which
questions are equivalent between the two frameworks so that answers for
those can be copied across. Anything not whitelisted will be assumed to
not be compatible (either the question is no longer asked in the new
framework, or the wording has materially changed and it would be
inappropriate to copy the data across).

## Related PR
https://github.com/alphagov/digitalmarketplace-content-loader/pull/56


Wynndow:
Well, apart from that failing test! Looks like a schema needs updating.

PR #516, opened 2018-04-16
Update questions to copy

samuelhwilliams:
## Summary
Some questions specify an ID to use in place of the filename. We had
incorrectly stored the filename in all instances rather than the ID,
where it was given. This corrects that so the data is correctly copied
across for all questions that are equivalent.

Bumps version to 11.5.2

### DIFF
--- a/frameworks/g-cloud-10/metadata/copy_services.yml
+++ b/frameworks/g-cloud-10/metadata/copy_services.yml
@@ -44,8 +44,7 @@ questions_to_copy:
   - dataProtectionWithinNetwork
   - dataProtectionWithinNetworkOther
   - dataSanitisation
-  - dataSanitisationTypeHosting
-  - dataSanitisationTypeSoftware
+  - dataSanitisationType
   - dataStorageAndProcessing
   - dataStorageAndProcessingLocations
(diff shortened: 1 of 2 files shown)


frameworks/g-cloud-10/metadata/copy_services.yml
```diff
@@ -8,9 +8,8 @@ questions_to_copy:
   - APIAutomationToolsOther
   - APIDocumentation
   - APIDocumentationFormats
-  - APIHosting
```
Wynndow: The yaml file for APIHosting and APISoftware don't specify an ID, they just have the name API.
Is the question identified by the name if no ID is specified? Is the filename not used at all?

samuelhwilliams: Yep, you're right. Was just running and testing it locally but you beat me to it. APIHosting/APISoftware are the keys used.

Wynndow: OK, so is the logic that is uses the filename, unless an ID is specified?

samuelhwilliams: Yep

PR #525, opened 2018-06-25
Add g9 following_framework meta and saved search temp messages

Wynndow:
Part of [this trello ticket.](https://trello.com/c/LjbF4cLc)

These are messages we need to show to buyers to inform them of how the
handover between framework iterations will work in relation to their
saved searches.

There is logic in the buyer frontend to decide which messages to use.
The variables are also populated there.


tests/test_metadata.py
```diff
@@ -5,18 +5,29 @@
 
 
 def test_copying_services_metadata():
-    """Make sure that every file/question listed under copying-services.questions_whitelist points to a valid
-    question file, and that the referenced historical framework exists."""
-    copying_services_filenames = glob.glob('frameworks/*/metadata/copying-services.yml')
+    """Make sure that every key listed under copy_services.questions_to_copy points to a valid
+    question file, id, or field and that the referenced historical framework exists."""
+    copying_services_filenames = glob.glob('frameworks/*/metadata/copy_services.yml')
```
samuelhwilliams: Should add an assertion here that `len(copying_services_filenames)` is > 0 or something.

PR #526, opened 2018-06-26
Update `generate-search-config` for G10

Wynndow:
The version of the frameworks repo is no longer defined in
`VERSION.txt`, instead it's in `package.json`

This also update the G10 search_mapping to have a `max_result_window` of
25000. This is because we have 24950 ish services to index.

### DIFF
--- a/frameworks/g-cloud-10/search_mappings/services.json
+++ b/frameworks/g-cloud-10/search_mappings/services.json
@@ -1,7 +1,7 @@
 {
   "settings": {
     "index": {
-      "max_result_window": 20000
+      "max_result_window": 50000
     },
     "analysis": {
       "analyzer": {
--- a/scripts/generate-search-config.py
+++ b/scripts/generate-search-config.py
@@ -26,6 +26,7 @@ Usage:
 """
 import os
 import sys
+import json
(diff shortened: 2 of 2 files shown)


frameworks/g-cloud-10/search_mappings/services.json
```diff
@@ -1,7 +1,7 @@
 {
   "settings": {
     "index": {
-      "max_result_window": 20000
+      "max_result_window": 25000
```
galund: I'd potentially suggest being a bit more generous. If we double instead, we'll probably not have to look at this again next time :-)

Wynndow: Yep, seems reasonable!

frameworks/g-cloud-10/search_mappings/services.json
```diff
@@ -1,7 +1,7 @@
 {
   "settings": {
     "index": {
-      "max_result_window": 20000
+      "max_result_window": 50000
```
risicle: You crazy fool.

Wynndow: I'm completely out of control.

PR #541, opened 2018-09-28
Fix DOS3 status search filters

Wynndow:
The way the open/closed filters was changed for DOS2 back in March, [see here](https://github.com/alphagov/digitalmarketplace-frameworks/commit/93a66aa7db3b522ac24d5ebecc9fee01708d698b). The change wasn't applied to DOS3 however, which causes weirdness on the search page. 

This catches DOS3 up.

There will be a PR on the buyer frontend once this is merged to pull these changes in.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-3/manifests/briefs_search_filters.yml
+++ b/frameworks/digital-outcomes-and-specialists-3/manifests/briefs_search_filters.yml
@@ -1,7 +1,7 @@
 -
   name: Status
   questions:
-    - status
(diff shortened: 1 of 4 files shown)


Wynndow:
@katstevens Yep, I think so. We cloned DOS2 to DOS3 in Jan I think, and this change was made to DOS2 in March. I've diffed the two directories in the repo to check for any other changes. This is the only thing that matters.

@lfdebrux I think there's a bit more info in the commit I linked to. Couldn't give you the main why though.

PR #544, opened 2019-01-24
Clone G11 from G10

katstevens:
Trello: https://trello.com/c/gJ6QDRRO/331-clone-g11-content-in-frameworks-repo

A basic clone of the G10 content, with a find-and-replace on `G-Cloud 10` -> `G-Cloud 11`. Everything in here will eventually be checked by a content designer. All dates, links etc are fictional and will be changed later.

Only a minor bump as the addition of this content shouldn't be a breaking change in itself.


lfdebrux:
Should the version bump come after the find and replace?

katstevens:
I don't think it really matters - the release is going to get tagged when the whole thing is merged...

katstevens:
Looking back at the last clone we did, I'm going to add a commit to put the dates into the far future to make sure we change them later.

katstevens:
Also added in `following_framework` - see commit message for details.

PR #545, opened 2019-01-29
G11 lot validation schema config

katstevens:
Forgot to add this config in when copying the G11 content.

It's used in `scripts/generate-validation-schemas.py`, which have already been run to generate schemas for the API (https://github.com/alphagov/digitalmarketplace-api/pull/861).


lfdebrux:
Woops I missed that in review!

PR #547, opened 2019-02-26
Update function and tests for JSON schema V7

jonodrew:
These changes update the number_property method and associated tests. These changes are a part of updating our JSON schema to version 7, which expects exclusiveMaximum to be a number rather than a boolean.

In updating the `number_property` function I found the `not limits.get('integer_only')` pattern confusing, and replaced it with a variable that explains what we're actually checking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 Records breaking changes from major version bumps
 
+## 14.0.0
+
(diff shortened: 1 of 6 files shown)


katstevens:
Will this count as a breaking change? If so can we add something to the changelog? We'll need a bump to the version in package.json anyway...

jonodrew:
Needs a changelog note - `requires jsonschema 3.x` and a version bump

PR #550, opened 2019-02-28
G11 service updates

katstevens:
Trello: https://trello.com/c/x5GVLSkq/384-implement-changes-for-legislation-changes-in-application-questions

Service question updates from G11.

Also minor declaration tweaks following on from v13.8.0.

Used to generate schemas for https://github.com/alphagov/digitalmarketplace-api/pull/903.

### DIFF
--- a/frameworks/g-cloud-11/manifests/edit_submission.yml
+++ b/frameworks/g-cloud-11/manifests/edit_submission.yml
@@ -245,7 +245,7 @@
   editable: false
   edit_questions: true
   questions:
-    - energyEfficientDatacentres
+    - multiqEnergyEfficiency
 
 - name: Pricing
   editable: false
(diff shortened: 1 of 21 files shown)


lfdebrux:
Does [`serviceInterfaceAccessibility.yml`](https://github.com/alphagov/digitalmarketplace-frameworks/blob/e265644f9133335cf28f1f6d74929e56718bfe04/frameworks/g-cloud-11/questions/services/serviceInterfaceAccessibility.yml#L14-L21) need to be updated to WCAG 2.1?

katstevens:
It should be fine, but I wanted to reword one of the commits anyway so have re-ordered.

katstevens:
Re: `serviceInterfaceAccessibility`, this change was missing from the doc but I think it is needed, yes. Good spot!

PR #552, opened 2019-03-05
Remove deprecated field `include_in_all`

lfdebrux:
Ticket: https://trello.com/c/uTA0PcGF

`include_in_all` is deprecated in Elasticsearch 5.6 and has been [removed][1] in Elasticsearch 6.0.

We were using `include_in_all: false` for the serviceIdHash as we didn't want people to be able to search by ID hash, but this option is now no longer available.

I'm currently looking into whether this will actually affect our search; I'm not sure if we ever search `_all` directly or indirectly. If we do we may have to decide whether we care about the hash being searchable because as far as I know there is no easy way to replicate the `include_in_all: false` behaviour in Elasticsearch 6.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-6.0.html#_the_literal_include_in_all_literal_mapping_parameter_is_now_disallowed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Records breaking changes from major version bumps
 
+## 15.0.0
+
+Removed the `include_in_all` parameter from the `serviceIdHash` field in the
+search mappings, as `_all` has been deprecated in Elasticsearch 5.6 and will be
+removed in Elasticsearch 6. Consumers of the Elasticsearch search mappings will
+now need to ensure that searches do not include the `serviceIdHash` field to
+remain compliant.
+
 ## 14.0.0
 
 Updated the `generate-validation-schemas` and `generate-assessment-schemas` scripts. These scripts now reflect the fact
--- a/frameworks/g-cloud-10/search_mappings/services.json
+++ b/frameworks/g-cloud-10/search_mappings/services.json
@@ -69,8 +69,7 @@
           "type": "keyword"
         },
         "sortonly_serviceIdHash": {
-          "type": "keyword",
-          "include_in_all": false
+          "type": "keyword"
         },
         "dmagg_lot": {
           "type": "keyword"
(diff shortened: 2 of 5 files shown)


katstevens:
I think we added the `service_id_hash ` field in to allow consistent search ordering: https://github.com/alphagov/digitalmarketplace-frameworks/pull/475

We certainly don't want to expose it anywhere (non-compliant searching for suppliers ahoy), so definitely worth checking if this would be a problem if used with our 5.x-created index.

Once these changes are merged, we'd have to regenerate the schemas for the search API, and potentially reindex all the G10 services with it. This would need to be factored into any migration plan.

(Given we're unlikely to bother ever regenerating the G9 service index we don't have to update that file.)

katstevens:
Oh this would need a version bump too.

lfdebrux:
> Oh this would need a version bump too.

What kind? 🤔

lfdebrux:
Marking this as WIP while we try and decide how to sort out the issue of how to avoid serviceIdHash being searchable, feel free to leave comments though.

>> Oh this would need a version bump too.

> What kind? 🤔

What we decide to do about searchIdHash will probably also affect what kind of version bump we need, so leaving this as an open question for now.

lfdebrux:
It looks like our search-api has already headed off this problem; queries that use `simple_query_string` (we don't use `query_string`) always have their [fields specified][1], and there is a [test][2] to make sure that `serviceIdHash` is not in the list of fields.

[1]: https://github.com/alphagov/digitalmarketplace-search-api/blob/master/app/main/services/query_builder.py#L119-L122
[2]: https://github.com/alphagov/digitalmarketplace-search-api/blob/master/tests/app/services/test_query_builder.py#L177-L184

lfdebrux:
> > Oh this would need a version bump too.

> What kind? 🤔

I decided to go for a major version bump, mainly because I felt the change needed to be documented in the changelog. It shouldn't actually affect anybody who pulls in the repo though...

risicle:
Yeah this _should_ be fine, I'd just make sure services aren't searchable by id-hash.

If they were actually, it wouldn't be critical because of the unlikelihood of someone typing a full hash - the problem would be if it matched against id-hash fragments or prefixes, and I don't think that's possible with `keyword` fields.

PR #555, opened 2019-03-08
Script to export declaration/service questions for a framework

katstevens:
We do this for each framework iteration, but after hours of searching I couldn't find the existing script. 

Provide either `--lot-slug` for service questions, or `--declaration` for declaration questions.

Contains some very rudimentary formatting to remove paragraph and list HTML tags from the content.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "digitalmarketplace-frameworks",
-  "version": "15.0.1",
+  "version": "15.1.0",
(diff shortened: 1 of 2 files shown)


katstevens:
Thankfully Rob's found the existing one, which even has tests 👍 https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/generate-questions-csv.py

PR #556, opened 2019-03-08
Delete old dos questions script

katstevens:
Not even a breaking change, as this DOS 1 script hasn't worked for ages and immediately `exit(1)`s anyway.


benvand:
lol 👋

PR #559, opened 2019-03-27
Metadata: list of services questions to scan for bad words

katstevens:
Trello: https://trello.com/c/Q0tQI9xL/439-spike-scan-g-cloud-services-for-bad-words-script-can-we-run-this-against-draft-services

Creates a new metadata list of service questions with free text inputs, that we will need to scan for bad/non-compliant words.

The `scan-g-cloud-services-for-bad-words` script (see https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/scan-g-cloud-services-for-bad-words.py) currently uses a hardcoded list of service questions from all recent frameworks (G7-G10). The script only ever scans services for one framework at a time, however, so we might as well start being framework specific now. See https://github.com/alphagov/digitalmarketplace-frameworks/tree/master/frameworks/g-cloud-11/questions/services for a list of G11 service questions.

The script itself will be refactored to use this list separately.


risicle:
'td be nice if we could have the tests assert that all of these field names actually exist.

katstevens:
@risicle that is an excellent idea.

katstevens:
Test now added, along with some schema validation (and a typo fix that the test caught!)

lfdebrux:
> Creates a new metadata list of service questions with free text inputs, that we will need to scan for bad/non-compliant words.

Did you create this file by hand or using a script? It would be good if that information were in the commit message?

tests/test_metadata.py
```diff
@@ -32,3 +32,33 @@ def test_copying_services_metadata():
...
+                    continue
+                with open(f'frameworks/g-cloud-11/questions/services/{filename}') as f:
+                    contents = yaml.safe_load(f)
+                    if contents.get('id'):
+                        names_ids_fields.add(contents['id'])
+                    if contents.get('fields'):
+                        names_ids_fields.update([v for v in contents.get('fields').values()])
+                    names_ids_fields.add(filename[:-4])
+
+            for question in copying_services_data['service_questions']:
```
lfdebrux: Is this for loop meant to be in the `for service_question_filename` loop?

katstevens: Some c+p from the test above has sneaked in. Will push a commit to fix.

katstevens:
@lfdebrux updated with the suggestions, thanks! I'll update the manual for this once the script's merged.

PR #564, opened 2019-04-29
Remove unnecessary clause from G11 declaration question

katstevens:
Trello: https://trello.com/c/1YUXLv7P/473-remove-unnecessary-clause-from-declaration-question

CCS and the G11 team agree that the following answer needs to be changed:

`Yes - your organisation has, or will have in place, employer’s liability insurance of at least $5 million and you will provide certification before the framework is awarded.`

To:

`Yes - your organisation has, or will have in place, employer’s liability insurance of at least $5 million`

The suppliers don't need to provide certification (CCS don't ask for it, and no-one provided it for G10!). Removing the clause should not cause a disadvantage to any suppliers.


lfdebrux:
Does this change also require an update to the search-api manifest?

PR #567, opened 2019-05-02
default.nix: add glibcLocales for nixpkgs 19.03's benefit

risicle:
Sorry, boring Nix thing.


risicle:
Nope - in this case Nix supplies the `yarn` and `node` that assemble the `package.json` deps, in the same way it supplies the python that grabs all the `venv` stuff. Not ideal, but works.

PR #585, opened 2019-07-04
Fix new accessibility questions for specialists

katstevens:
Unfortunately it looks like sharing the same question across 7 specialist roles isn't going to work.

This PR:
- splits the `accessibleApplicationsSpecialists` question into separate questions for each role
- adds a human readable error message if the user answers 'no' to that question (will require a small API change, PR incoming, which will also include the resulting schema changes)


lfdebrux:
For reference, why doesn't it work to share the question?

katstevens:
@lfdebrux when I tested adding a new role, the answer I gave for it showed up for all of them 😹

lfdebrux:
> @lfdebrux when I tested adding a new role, the answer I gave for it showed up for all of them 😹

Ahhhhh hahaha

PR #588, opened 2019-07-11
dos4 briefs evaluationType* questions: remove before_summary_value, change content to make it clearer these are additional

risicle:
https://trello.com/c/S1kIwe05

Once brought through to the briefs frontend this goes from looking like

image

for digital outcomes in the initial state to 

image

with similar changes for digital specialists and UR participants.


risicle:
> Fab +1 Let's try it!

:goes-on-holiday:

PR #592, opened 2019-09-06
Remove Yarn in favour of npm

lfdebrux:
We do not need to depend on Yarn as npm is Nodes default package manager and we use npm to install yarn. npm does everything yarn does now.

This commit mainly updates the manual but it also replaces `.yarnrc` with `.npmrc`.

Trello ticket: https://trello.com/c/RrpjQDJi/278-replace-yarn-with-npm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,12 @@
+# Node stuff
+node_modules
+
+# Python stuff
 __pycache__
 *.pyc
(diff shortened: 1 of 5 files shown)


lfdebrux:
> Need to remove the two lock files in .gitignore and include this pr needs a package-lock.json file add ( `npm install`)

Done

PR #596, opened 2019-11-14
Remove customer benefits record email

lfdebrux:
Ticket: https://trello.com/c/nY41xU2U/105-modify-copy-and-change-url-for-customer-benefits-record-form

CCS no longer use this email address (customer benefit records are now submitted via Google Form, see #595), so we can remove this metadata from the G-Cloud framework.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 Records breaking changes from major version bumps
 
+## 17.0.0
+
+Removes the customer benefits record email from G-Cloud 11 manifest, as this is no longer used by CCS. This was only
+used in the G-Cloud Direct Award flow in the Buyer Frontend.
+
 ## 16.0.0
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Records breaking changes from major version bumps
 
+## 17.0.0
+
+Removes the customer benefits record email from G-Cloud 11 manifest, as this is no longer used by CCS.
```
katstevens: Pretty sure the email is only used in the Buyer FE's Direct Award flow templates? Let's say so here, so that people don't have to go digging all over again when they pull this into the other FE apps.

katstevens:
Ooo I always forget we have tests for the schemas!

lfdebrux:
> Ooo I always forget we have tests for the schemas!

Hmm, I guess I can just leave the property in the schema, there's probably no harm in it being there.

katstevens:
Ah I see - the old G9/G10 iterations are failing. Sure, let's leave it in.

PR #598, opened 2019-12-12
Helper script for cloning framework content from a previous iteration

katstevens:
Automates some of the initial steps here: https://alphagov.github.io/digitalmarketplace-manual/content-and-frameworks/adding-frameworks.html#id2

It's verbose/non-DRY in places, to try and keep visibility of the different files being edited.

Tested with G-Cloud 12 and DOS 5, but not exhaustively.

Scope for more unit tests to be added at a later date 😸

### DIFF
--- a/script_helpers/clone_helpers.py
+++ b/script_helpers/clone_helpers.py
@@ -36,7 +36,13 @@ class FrameworkContentCloner:
     def __init__(self, framework_family, iteration_number, launch_year):
         self._launch_year = launch_year
         self._new_fw_slug = f"{framework_family}-{iteration_number}"
-        self._previous_fw_slug = f"{framework_family}-{iteration_number - 1}"
+        if iteration_number == 2:
+            # Handle first iterations without numeric suffix (e.g. DOS 2 -> DOS)
+            self._previous_fw_slug = f"{framework_family}"
+        elif iteration_number < 2:
+            raise ValueError("Can't clone a framework on its first iteration")
+        else:
+            self._previous_fw_slug = f"{framework_family}-{iteration_number - 1}"
         self._following_fw_slug = f"{framework_family}-{iteration_number + 1}"
 
         self._previous_name = get_fw_name_from_slug(self._previous_fw_slug)
--- a/script_helpers/clone_helpers.py
+++ b/script_helpers/clone_helpers.py
@@ -1,3 +1,21 @@
+import os
+import sys
+import shutil
+import yaml
+
+
+METADATA_FILES = {
+    'copy_services': 'copy_services.yml',
+    'following_framework': 'following_framework.yml',
+    'bad_words': 'service_questions_to_scan_for_bad_words.yml'
+}
+
+DATES_FILES = {
+    'application_deadline': 'messages/homepage-sidebar.yml',
+    'clarification_deadline': 'questions/declaration/understandHowToAskQuestions.yml',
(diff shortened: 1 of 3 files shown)


script_helpers/clone_helpers.py
```diff
@@ -0,0 +1,152 @@
...
+    last_space_position = name.rindex(" ")
+    return name[:last_space_position] + "&nbsp;" + name[(last_space_position + 1):]
+
+
+class FrameworkContentCloner:
+
+    def __init__(self, framework_family, iteration_number, launch_year):
+        self._launch_year = launch_year
+        self._new_fw_slug = f"{framework_family}-{iteration_number}"
+        self._previous_fw_slug = f"{framework_family}-{iteration_number - 1}"
```
risicle: Not expecting to have to care about e.g. DOS1 omitting the number altogether, are we?

scripts/clone-latest-framework.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+import os
+import sys
+sys.path.insert(0, '.')
+from docopt import docopt
+from script_helpers.clone_helpers import FrameworkContentCloner
+
+
+if __name__ == '__main__':
+    current_dir = os.path.dirname(__file__)
```
risicle: Das ist nicht used, nein?

scripts/clone-latest-framework.py
```diff
@@ -0,0 +1,35 @@
...
+- Update metadata
+- Update hardcoded content of framework name in declaration questions 'G-Cloud 11', 'G-Cloud&nsbp;11'
+- Update hardcoded content of framework slug in urls, declaration
+- Set any dates to a point in the far-future e.g. 1 Jan 2525
+
+Usage:
+    clone-latest-framework.py --family=<framework_family> --iteration=<iteration_number> --launch-year=<launch_year>
+
+Example:
+    ./scripts/clone-latest-framework.py --family=g-cloud --iteration=12 --launch-year=2020
```
risicle: It's not been made clear here what `--launch-year` is for.

tests/test_clone_latest_framework.py
```diff
@@ -0,0 +1,27 @@
...
+    assert cloner._launch_year == 2020
+    assert cloner._new_fw_slug == 'g-cloud-12'
+    assert cloner._previous_fw_slug == 'g-cloud-11'
+    assert cloner._following_fw_slug == 'g-cloud-13'
+
+    assert cloner._previous_name == 'G-Cloud 11'
+    assert cloner._new_name == 'G-Cloud 12'
+    assert cloner._previous_nbsp_name == 'G-Cloud&nbsp;11'
+    assert cloner._escaped_previous_nbsp_name == 'G&#x2011;Cloud&nbsp;11'
+    assert cloner._new_nbsp_name == 'G-Cloud&nbsp;12'
```
risicle: Now a _purist_ would scream at you for testing a non-public interface.
(I'm not a purist though)

script_helpers/clone_helpers.py
katstevens: Good point!

tests/test_clone_latest_framework.py
katstevens: Hah, there are barely any tests as it is - I'm not removing these 😸

script_helpers/clone_helpers.py
katstevens: Have added a commit to handle `iteration_number = 1` as well.

PR #602, opened 2020-01-31
257-service-question-change-make-servicedefinitiondocumenturl-mandatory

benvand:
https://trello.com/c/VLjBsVoc/257-service-question-change-make-servicedefinitiondocumenturl-mandatory

Make service definition document mandatory

### DIFF
--- a/frameworks/g-cloud-12/questions/services/serviceDefinitionDocumentURL.yml
+++ b/frameworks/g-cloud-12/questions/services/serviceDefinitionDocumentURL.yml
@@ -5,7 +5,6 @@ hint: >
   This document will not be indexed by search on the Digital Marketplace. Use an Open Document Format (ODF) or PDF/A
   (eg .pdf, .odt). (Maximum file size 5MB.)
 
-optional: true
 depends:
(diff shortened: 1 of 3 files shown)


frameworks/g-cloud-11/questions/services/serviceDefinitionDocumentURL.yml
```diff
@@ -5,7 +5,7 @@ hint: >
   This document will not be indexed by search on the Digital Marketplace. Use an Open Document Format (ODF) or PDF/A
   (eg .pdf, .odt). (Maximum file size 5MB.)
 
-optional: true
+optional: false
```
katstevens: We can just remove the entire `optional` key, I think - as per the `termsAndConditionsDocumentURL` question

katstevens: We can also get rid of the `empty_message` key below.

benvand: Done 👍

PR #603, opened 2020-01-31
assessment schema improvements

risicle:
This is to fulfil https://trello.com/c/kbTzccXb

Firstly I discovered that the "definite pass" G11 declaration answer used in the tests _wasn't_ a definite pass - it had missed out on some field renames/updates/splits since G10, but hadn't detected the "wrongness" because the generated schemas didn't `require` answers for any fields. So the first thing I did was make all assessed questions _by default_ required. I needed to make it possible to override this in the question yaml because of `modernSlaveryReportingRequirements`, which is the one exception.

Interestingly, the purpose of this `required` setting is **only** to help guarantee we keep the content, the test's "definite pass" candidate and the supplier frontend's custom declaration code in sync. We don't actually rely on it to police the dependencies between the questions.

The actual problem in the trello card is solved by adding support for a framework to supply "extra" schema json. This allows us to support more complex rules which are hard to express otherwise and rules which relate to more than one question. The generated schema was also changed to put the generated constraints in a redundant `allOf` to make it easier and more reliable to add new conditions.

The conditions are added through deep merging, done using the `deepmerge` package. I considered using `jsonmerge` - a very interesting library and one we could transition to if ever necessary. But more complex than we need right now (I recommend taking a look at it though, it's quite cool).

I had trouble deciding whether to accept baseline and discretionary "extra schema" separately and so make the mechanism opaque to how we implement the baseline/discretionary split, but decided against it in the end because we already rely on having a separate sub-schema named `baseline` in normal usage.

### DIFF
--- a/frameworks/g-cloud-11/questions/declaration/modernSlaveryReportingRequirements.yml
+++ b/frameworks/g-cloud-11/questions/declaration/modernSlaveryReportingRequirements.yml
@@ -15,3 +15,5 @@ assessment:
   passIfIn:
     - True
   discretionary: True
+  # actually dependent, but it is not the assessment schema's responsibility to enforce that dependency
+  required: False
--- a/frameworks/g-cloud-12/assessment/declaration/extra.schema.json
+++ b/frameworks/g-cloud-12/assessment/declaration/extra.schema.json
@@ -1,0 +1,14 @@
+{
+    "definitions": {
+        "baseline": {
+            "allOf": [
+                {
+                    "anyOf": [
+                        {"properties": {"servicesHaveOrSupportCloudHostingCloudSoftware": {"enum": ["Yes"]}}},
+                        {"properties": {"servicesHaveOrSupportCloudSupport": {"enum": ["Yes"]}}}
+                    ]
+                }
+            ]
+        }
+    }
+}
--- a/schema_generator/assessment.py
+++ b/schema_generator/assessment.py
@@ -28,14 +28,12 @@ def generate_schema(framework_slug, question_set, manifest_name):
         elif "passIfIn" in question.get("assessment", {}):
             assessed_questions.append(question)
 
-    discretionary_properties = {
-        question.id: _enum_for_question(question)
-        for question in assessed_questions if question["assessment"].get("discretionary")
-    }
-    baseline_properties = {
-        question.id: _enum_for_question(question)
-        for question in assessed_questions if not question["assessment"].get("discretionary")
-    }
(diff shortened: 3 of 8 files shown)


schema_generator/assessment.py
```diff
@@ -28,29 +50,33 @@ def generate_schema(framework_slug, question_set, manifest_name):
...
         "definitions": {
             "baseline": {
                 "$schema": "http://json-schema.org/draft-04/schema#",
                 "title": "{} Declaration Assessment Schema (Baseline Schema)".format(framework_slug),
                 "type": "object",
-                "properties": baseline_properties,
+                "allOf": [
+                    {"properties": {q.id: _enum_for_question(q) for q in baseline_questions}},
+                ],
+                "required": sorted(q.id for q in baseline_questions if q["assessment"].get("required", True)),
```
katstevens: Have you managed to get this new schema to work with the `export-framework-results-reasons` script? When I try, the errors for the new `required` fields are output by the validator, but they have no `path` so the script (in its current form, at least) throws an error: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/export_framework_results_reasons.py#L22

frameworks/g-cloud-12/assessment/declaration/extra.schema.json
```diff
@@ -0,0 +1,14 @@
+{
```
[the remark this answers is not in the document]
risicle: Indeed it does.

frameworks/g-cloud-11/questions/declaration/modernSlaveryReportingRequirements.yml
```diff
@@ -15,3 +15,5 @@ assessment:
   passIfIn:
     - True
   discretionary: True
+  # actually dependent, but it is not the assessment schema's responsibility to enforce that dependency
+  required: False
```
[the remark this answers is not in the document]
risicle: - [ ] :+1:

schema_generator/assessment.py
risicle: Ugh I had a sneaking suspicion about that script...

I'll take a look - it will probably need some attention.

risicle: Is this error happening for you when the `required` constraint is being violated? Finding this hard to reproduce.

risicle: Handled in https://github.com/alphagov/digitalmarketplace-scripts/pull/465

PR #608, opened 2020-02-21
G12 file upload guidance

katstevens:
https://trello.com/c/Rtcc6oi6/316-add-link-to-guidance-on-making-accessible-pdfs-on-application-5-places

We want to provide suppliers with guidance on uploading accessible documents.

Paired with @NoraGDS on the content.

There are 6 questions that involve file uploads:

1) Modern Slavery Statement (for suppliers with < £36m turnover) 
2) Modern Slavery Statement (for suppliers with >= £36m turnover)
3) Service definition document
4) Terms and conditions document
5) SFIA rate card 
6) Pricing document

### Screenshot for 1 (Q2 is very similar):
pdf-guidance-modern-slavery

### Screenshot for 3 (Q4, 5 and 6 are very similar): 
pdf-guidance-service-def-doc

### DIFF
--- a/frameworks/g-cloud-12/questions/declaration/modernSlaveryReportingRequirements.yml
+++ b/frameworks/g-cloud-12/questions/declaration/modernSlaveryReportingRequirements.yml
@@ -1,8 +1,6 @@
 name: Modern slavery reporting requirements
 question: Does your organisation comply with the annual reporting requirements of section 54 of The Modern Slavery Act (2015)?
-
-hint: >
-  You must publish a slavery and human trafficking statement to comply with the act. Read the Home Office guidance for information about <a href="https://www.gov.uk/government/publications/transparency-in-supply-chains-a-practical-guide" target="_blank" rel="noopener noreferrer">how to write a statement and what information it must include (link opens in a new tab)</a>.
+hint: You must publish a slavery and human trafficking statement to comply with the act.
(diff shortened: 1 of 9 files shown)


risicle:
I'd love if we could say PDF/A-1, because we certainly shouldn't be accepting e.g. PDF/A-3

katstevens:
> I'd love if we could say PDF/A-1, because we certainly shouldn't be accepting e.g. PDF/A-3

@NoraGDS - what do you think? The linked guidance doesn't mention what PDF/A-1 is anywhere...

NoraGDS:
while I can see it would help with embedding files of any type - it may be confusing because the GDS guidance doesn't refer to the levels of PDF/A or how to produce them. Suggest for this framework we keep it consistent with the guidance.

katstevens:
I've created a tech debt card to capture this for DOS5: https://trello.com/c/EijTv0m8/1287-provide-more-guidance-on-what-constitutes-a-pdf-a-1-opposed-to-pdf-a-2-a-3-etc

@risicle Are you ok with that for now?

risicle:
Well, if we're going to start rejecting PDFs, we should at least let them know accurately what we *will* accept. Just saying PDF/A gives us much less grounds to reject a PDF/A-3 that they then give us.

PR #599, opened 2020-01-23
256-clone-framework-content

benvand:
https://trello.com/c/ePgvYRGr/256-clone-framework-content-just-clone-it-with-kats-script

* Clone framework g-cloud-11 for g-cloud-12, updating dates, following framework, etc. (used scripts/clone-latest-framework.py)

  * Unblocks https://trello.com/c/Hbojps0X/265-check-all-dates-in-content-for-g12

PR #609, opened 2020-02-25
Additional copy_services fields 

katstevens:
https://trello.com/c/OZT4aPjn/367-copying-services-some-missing-fields

We reinstated some of the fields during the initial clone of the G12 content (https://github.com/alphagov/digitalmarketplace-frameworks/pull/599), but some got missed. This caused a bug, because we reintroduced `energyEfficientDatacentres`, but not `energyEfficientDatacentresDescription` (a dependent question).

There are also a few fields from G9 that never made it into the G11 copy list! See ticket for details.


katstevens:
I think the Travis build stalled during the github issue the other day (the builds all passed, but the results never got back to this PR). I've restarted it.

PR #621, opened 2020-05-19
remove ONS link from DOS4 framework

gidsg:
This removes the ONS link as per https://trello.com/c/6b806g3r/40-remove-map-pdf-from-dos-brief-where-supplier-will-work
(I also removed it from the supplier questions, as it is assumed we would not want these links either in future iterations of the framework)
Screenshot 2020-05-19 at 15 09 40


katstevens:
🔥I love deleting things. 

This will need the version number updating so we can pull it into the apps - for this repo it's in [package.json](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/package.json).

PR #622, opened 2020-05-28
Update error messages for DOS supplier questions and answers

lfdebrux:
Make the error message content more in line with GOV.UK Design System guidance.


gidsg:
Looks good to me, is it worth @NoraGDS taking a look if she hasn't already?

lfdebrux:
> Looks good to me, is it worth @NoraGDS taking a look if she hasn't already?

I think it's okay but it's not urgent so happy to wait

NoraGDS:
happy with the content amends

PR #623, opened 2020-05-29
Tweak DOS4 brief questions

lfdebrux:
Ticket: https://trello.com/c/U2OsGeCq/26-2-high-level-content-review-of-pages-in-create-a-dos-opportunity-journey

Content updates from @NoraGDS.


frameworks/digital-outcomes-and-specialists-4/manifests/display_brief.yml
```diff
@@ -40,7 +40,8 @@
   name: Work setup
   questions:
     - workplaceAddress
-    - workingArrangements
+    - workingArrangementsOutcomes
+    - workingArrangementsSpecialists
```
katstevens, line 44: I don't think we can split out the question in this way. There are going to be a bunch of existing DOS4 briefs with the old `workingArrangements` in the JSON data that will no longer match up to the questions in `display_brief` and `edit_brief`. We normally make changes like this per framework iteration, to make sure the manifest/questions are in line with the data.

lfdebrux: I haven't seen any issues testing locally, I believe this is because the content loader identifies questions by id rather than filename

https://github.com/alphagov/digitalmarketplace-content-loader/blob/master/dmcontent/questions.py#L670-L688

These questions didn't have an id previously, which means that the filename was used instead

https://github.com/alphagov/digitalmarketplace-content-loader/blob/0812d0a931c1b93f00317eff862cf9a0c0d3f55f/dmcontent/content_loader.py#L664-L671

But since I added an id equal to the filename before renaming, it all seems to work out.

lfdebrux: I'm happy to take the change out though if you think it's too risky.

katstevens: Were you able to edit that field on the brief once you'd created it? The API brief schema has `workingArrangements` as a required field so I'm surprised it worked...

lfdebrux: I was. I tested this by 1) creating a specialists brief using the master version of briefs frontend, then 2) cd-ing to the briefs frontend, 3) deleting the folder `app/content`, 4) installing the local version of the frameworks repo with `npm install ../digitalmarketplace-frameworks`, 5) rebuilding the briefs frontend with dmrunner, and then 6) visiting the same new brief and editing the field. There was no difference in behaviour that I could see between steps 1 and 6

katstevens: 🤔  When I do an API `POST` update to an old draft brief with payload `{"briefs": {"workingArrangementsSpecialists" : "whatever" }}`, I get the following:

```
{
    "error": {
        "_form": [
            "Additional properties are not allowed ('workingArrangementsSpecialists' was unexpected)"
        ]
    }
}
```

lfdebrux: I don't think that's what the briefs frontend POSTs though, because the question id is `workingArrangements`?

frameworks/digital-outcomes-and-specialists-4/questions/briefs/evaluationTypeSpecialists.yml
```diff
@@ -15,22 +15,22 @@ optional: true
 options:
   - label: "Reference"
     description: >
-       A written reference is a good way of confirming whether a specialist has done the work they’ve said they’ve
-       done. You should always ask for references that relate to a specific timeframe, eg work they’ve done in the
-       last 2 years.
+      A written reference is a good way of confirming whether a specialist has done the work they’ve said they’ve
+      done. You should always ask for references that relate to a specific timeframe, for example: work they’ve done
+      in the last 2 years.
```
NoraGDS, line 20: Please replace 'for example: with: such as

frameworks/digital-outcomes-and-specialists-4/questions/briefs/evaluationTypeSpecialists.yml
```diff
@@ -6,7 +6,7 @@ question_advice: |
 
   Choose additional ways to assess specialists.
 
-  You can only use the methods you choose here, eg if you don’t select an interview, you won’t be able to interview specialists.
+  You can only use the methods you choose here, for example: if you don’t select an interview, you won’t be able to interview specialists.
```
NoraGDS, line 9: Please replace 'for example:' with: so

frameworks/digital-outcomes-and-specialists-4/manifests/display_brief.yml
katstevens: Ah right, I'm with you now. I guess it will work, but I would really advise against changing the structure of questions etc mid-framework cycle. Whenever we do it seems there's always a bit of hardcoded horror lurking somewhere to trip us up. Given the content change is so small, if we can leave this one out, I would.

frameworks/digital-outcomes-and-specialists-4/questions/briefs/evaluationTypeSpecialists.yml
lfdebrux: Done

frameworks/digital-outcomes-and-specialists-4/questions/briefs/evaluationTypeSpecialists.yml
lfdebrux: Done

frameworks/digital-outcomes-and-specialists-4/questions/briefs/evaluationTypeSpecialists.yml
lfdebrux: Done

frameworks/digital-outcomes-and-specialists-4/manifests/display_brief.yml
lfdebrux: No worries. It was good to go through this questioning so I could improve my understanding of it.

@NoraGDS are you happy for the question advice for the working arrangements question to stay as "Describe how you want to work with the individual or team." everywhere (even where it can't be an individual)?

NoraGDS: Yes, fine to leave it as it is due to the reasons described. Will make a note for a future DOS5 board that we look at this before the next framework iteration.

lfdebrux:
I've updated this based on further discussion on the [journeys spreadsheet](https://docs.google.com/spreadsheets/d/1ZGo9L-0DzMWPJwrCAAEGHR1QNr1AtfgveoMoU07fpKQ/edit#gid=1515477901)

PR #624, opened 2020-06-22
Use headings instead of bold text

katstevens:
https://trello.com/c/X4vKC3Bv/112-2-apply-govuk-frontend-styles-to-frameworks-content-for-the-create-a-dos-opportunity-journey

We shouldn't really be using bold text in the places that we are, really - the Design System says "[use bold sparingly](https://design-system.service.gov.uk/styles/typography/)".

I've converted the examples below to headings, as that will give the same font size/weight as the other headings on the page (which is what we have currently anyway).


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally
- [ ] run functional tests locally

PR #631, opened 2020-09-15
Revert accreditation date field to 100 char limit

bjgill:
Trello: https://trello.com/c/lwJW7PvN/288-2-create-dos5-service-schemas

The limit is 100 in digitalmarketplace-api (https://github.com/alphagov/digitalmarketplace-api/blob/f71d124/json_schemas/services-g-cloud-12-cloud-software.json#L3067), but 50 in this repo.

It was originally 100, but was changed in 62d3375f1baea0bbf8b3dbb1c5b1876694868adc. This change to 50 looks like a copy-paste error to me, so reverting to 100 is the right thing to do.

I don't know, but would be surprised if increasing the limit has any negative effects.


lfdebrux (approved):
This makes sense to me. Thanks for going the extra mile with the detailed investigation.

PR #632, opened 2020-09-15
Assume less in tests

bjgill:
The [`integers`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.integers) strategy allows us to specify min and max values (inclusively). Thus, we can use those instead of assuming. This should make those tests slightly faster, and reduce the risk of Hypothesis failing altogether because it can't find enough good examples.

Might reduce the risk of future [build failures](https://travis-ci.com/github/alphagov/digitalmarketplace-frameworks/builds/184405782)

### DIFF
--- a/tests/test_generate_validation_schema.py
+++ b/tests/test_generate_validation_schema.py
@@ -141,11 +141,9 @@ def test_uri_property(id_name):
         "format": "uri"}}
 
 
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
Looks good!

PR #638, opened 2020-09-29
Pull latest version of content loader and utils

bjgill:
Supplier frontend is already using v7.22.0, so it should be safe to update the frameworks repo as well.

Modify test to be able to strip out more complex paragraph tags.

### DIFF
--- a/tests/test_content.py
+++ b/tests/test_content.py
@@ -1,4 +1,5 @@
 import glob
+import re
 
 import mock
 import pytest
@@ -107,7 +108,8 @@ def test_render_question(framework, question_id, question_set, question):
 
 
 def strip_paragraph_wrapper(value):
-    if value.startswith('<p>'):
-        value = value[3:-4]
+    match = re.fullmatch(r"^(<p.*?>)(.+?)(</p>)", value)
+    if match:
+        return match.group(2)
 
     return value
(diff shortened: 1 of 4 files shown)


tests/test_content.py
```diff
@@ -107,7 +108,8 @@ def test_render_question(framework, question_id, question_set, question):
 
 
 def strip_paragraph_wrapper(value):
-    if value.startswith('<p>'):
-        value = value[3:-4]
+    match = re.fullmatch(r"^(<p.*?>)(.+?)(</p>)", value)
+    if match:
+        return match.group(2)
```
lfdebrux, line 113: Could you add a commit message explaining why this change is needed?

lfdebrux (approved):
Great! ➕

tests/test_content.py
bjgill: :+1: 

    Modify test to be able to strip out more complex paragraph tags. We need to do this because the previous logic was only able to strip '<p>'. However, as of https://github.com/alphagov/digitalmarketplace-content-loader/pull/77, we get more complex paragraph tags such as '<p class="govuk-body">'. The new regex should be able to detect these more complex paragraph tags and allow us to strip them.

bjgill:
Rebased (with better commit message) and bumped minor version. Just awaiting Travis now

PR #645, opened 2020-10-07
Add GOV.UK Design System styles to HTML elements in DOS5 content

lfdebrux:
Ticket: https://trello.com/c/dz7qu083/425-2-ensure-hardcoded-html-in-dos-framework-application-content-has-appropriate-design-system-styles

Makes sure we're using [GOV.UK Design System styles](https://design-system.service.gov.uk/styles/typography/) in DOS5 application content. Mostly making sure that links have the proper class. We should look at if there's some way we can make this easier by having the content loader add all the open in new tab attributes, instead of having to copy and paste for each link.

I also updated the 'become a supplier' message for G12 to make sure the content on that page is consistent.

There are still a few HTML elements in DOS5 brief responses content which I didn't update because it's hard to test that at this point.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-5/manifests/declaration.yml
+++ b/frameworks/digital-outcomes-and-specialists-5/manifests/declaration.yml
@@ -4,7 +4,7 @@
   editable: True
   prefill: False
   description: |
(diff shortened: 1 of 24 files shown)


bjgill (approved):
Looks good to me.

PR #646, opened 2020-10-07
Enable prefill for DOS 5 declaration questions where possible

bjgill:
Trello: https://trello.com/c/5alXnNVP/429-stop-giving-users-the-option-to-prefill-the-declaration-or-turn-on-prefill-for-something

Prefill helps suppliers fill in their declarations faster by reusing their answers from previous frameworks. We need to disable prefill for questions that have significantly changed.

However, for DOS 5 many of the questions have not changed significantly. We can also use clever tricks to enable prefill for a set of questions and only disable it for the particular questions that actually changed.

See the individual commit messages for an explanation of why each change is safe to make.

Also, for those questions that have prefill disabled, we can take this opportunity to give them values for their answers. This decouples the content from the logic and should make it easier to change the content in future without disrupting prefill.

I've tested this all manually locally and confirmed that the pre-fill was working as expected.


frameworks/digital-outcomes-and-specialists-5/questions/declaration/GAAR.yml
```diff
@@ -1,4 +1,5 @@
 name: Why a tax return was incorrect
+id: taxReturns
```
lfdebrux, line 2: This is a big improvement over GAAR, whatever that is. Could we go one further and call it `incorrectTaxReturns`?

Also, is there any reason not to rename the file, rather than adding an explicit id?

frameworks/digital-outcomes-and-specialists-5/questions/declaration/safeguardOfficialInformation.yml
```diff
@@ -1,4 +1,5 @@
 name: Safeguarding official information
+id: safeguardOfficialGovernmentInformation
```
lfdebrux, line 2: Why not `safeguardingOfficialInformation`?

frameworks/digital-outcomes-and-specialists-5/manifests/declaration.yml
```diff
@@ -141,7 +141,7 @@
   name: Application accuracy
   slug: application-accuracy
   editable: True
-  prefill: False
+  prefill: True
```
lfdebrux, line 144: This had prefill disabled in DOS4 and DOS3 even though the question didn't change, I think because it doesn't make sense to prefill this confirmation.

frameworks/digital-outcomes-and-specialists-5/questions/declaration/GAAR.yml
bjgill: Good idea. Renamed.

(General Anti-Abuse Rule, FYI)

frameworks/digital-outcomes-and-specialists-5/questions/declaration/employersInsurance.yml
```diff
@@ -8,9 +9,12 @@ hint: |
 type: radios
 options:
   - label: "Yes, your organisation has, or will have in place, employer’s liability insurance of at least £5 million and you will provide certification before the framework is awarded"
+    value: insurance
   - label: "No, your organisation does not have, and will not have in place, employer’s liability insurance of at least £5 million before the framework is awarded"
+    value: no-insurance
   - label: "Not applicable, your organisation does not need employer’s liability insurance because your organisation employs only the owner or close family members"
+    value: not-applicable
```
lfdebrux, line 16: This is a good idea.

frameworks/digital-outcomes-and-specialists-5/questions/declaration/safeguardOfficialInformation.yml
bjgill: The purpose of this change is to change the ID of this question so that it is not pre-filled.

`safeguardingOfficialInformation` is the current ID as generated from the file name (AFAIK). So we need to set the ID to something different in order to avoid prefill.

frameworks/digital-outcomes-and-specialists-5/manifests/declaration.yml
bjgill: Good point. I'll add a comment to help us avoid this mistake next year.

lfdebrux (approved):

PR #649, opened 2020-10-12
Exclude fields from DOS5 copying

gidsg:
https://trello.com/c/rYwa42rs/1954-error-messages-dos5
We were erroneously copying some questions from DOS4 which had been deleted in DOS5 which caused users to experience `400` errors when attempting to save services.


lfdebrux (approved):
Looks good to me

PR #650, opened 2020-10-12
Refactor metadata tests

lfdebrux:
Refactors the tests for copy service framework metadata, and adds a new test that checks if things are missing from questions_to_exclude that should be present. This should reduce the chances of the issues we saw in https://trello.com/c/rYwa42rs/1954-error-messages-dos5.


gidsg (approved):
nice refactor! 👍

PR #654, opened 2020-11-19
Switch pass criterion

bjgill:
We flipped the meaning of the question in ff63b37513f88f611653a158cef32174adfcfafc, but forgot to update the assessment criterion.

I've checked and this is the only one we need to fix


gidsg (approved):
Good spot! 👍

lfdebrux (approved):

PR #656, opened 2020-12-04
Update error messages for brief response journey

domoscargin:
https://trello.com/c/x5aacZhG/282-1-replace-validation-banner-with-govuk-frontend-error-summary-component-in-the-brief-responses-frontend

Updating the error messages for applying to a brief.

There's two potentially tricky cases I'm not sure the best approach for.

### How to make these distinguishable?
On skills evidence questions, we can either go with something generic like:

"Enter evidence for this essential skill"

Or something dynamic like:

"Enter evidence for {{ item }}"

Or use the advice on the ticket and stick with our format of listing the questions:

There is a problem:
Question 1
Question 2

With the first approach, a user might get multiple messages that are exactly the same. With the second, you can potentially get a very long string, like "Enter evidence for Able to operate a forklift truck at the same time as doing some juggling and yelling loudly". With the third, there's no action for the user.

## Combined yes/no and evidence textarea
On the nice to have requirements, first a user has to select yes or no. If they select yes, a textarea is revealed. If they leave that textarea blank, the current message is:

"You must provide evidence if you answer ‘yes’ to this question."

Which might be OK, but might be a bit unclear - the error is on the evidence textarea, but "this question" refers to the yes/no radio.


NoraGDS:
<b>Combined yes/no and evidence textarea</b>
The position of error message looks clear on this because it's beneath description of the evidence rather than directly below the radio button. To avoid any confusion with the question, an alternative wording is: 'You must provide evidence of this skill or experience'; or to avoid having the same error message multiple times: 'You must provide evidence of {{item}}'.


<b>Evidence textareas</b>
The dynamic option of "Enter evidence for {{ item }}" is the one that best meets the Design System pattern. It avoids having the same error message with no context for the user, even if the criteria makes it long in some cases.

lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally
- [ ] run functional tests locally

frameworks/digital-outcomes-and-specialists-4/questions/brief-responses/dayRate.yml
```diff
@@ -11,12 +11,12 @@ depends:
...
-    message: 'You need to answer this question.'
+    message: 'Enter a day rate'
   - name: not_money_format
     field: dayRate
-    message: "Day rate must be in numbers and decimal points only, for example 99.95."
+    message: "Day rate must be in numbers and decimal points only, for example 99.95"
   - name: max_less_than_min
     field: dayRate
-    message: "This can't be more than the maximum day rate you've already provided."
+    message: "Day rate cannot be more than the maximum day rate you've already provided"
```
lfdebrux, line 20: I think we should be using a curly apostrophe here

```suggestion
    message: "Day rate cannot be more than the maximum day rate you’ve already provided"
```

frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/dayRate.yml
```diff
@@ -11,12 +11,12 @@ depends:
...
-    message: 'You need to answer this question.'
+    message: 'Enter a day rate'
   - name: not_money_format
     field: dayRate
-    message: "Day rate must be in numbers and decimal points only, for example 99.95."
+    message: "Day rate must be in numbers and decimal points only, for example 99.95"
   - name: max_less_than_min
     field: dayRate
-    message: "This cannot be more than the maximum day rate you've already provided."
+    message: "Day rate cannot be more than the maximum day rate you've already provided"
```
lfdebrux, line 20: I think we should be using a curly apostrophe here

```suggestion
    message: "Day rate cannot be more than the maximum day rate you’ve already provided"
```

lfdebrux:
The dynamic question errors looks really cool, my only nitpick is that the first letter of {item} is capitalised:

- Enter evidence for Boil kettle - Enter evidence for Taste tea

Is this a problem? Do we want to try and fix it?

domoscargin:
> The dynamic question errors looks really cool, my only nitpick is that the first letter of {item} is capitalised:
> 
> - Enter evidence for Boil kettle - Enter evidence for Taste tea
> 
> Is this a problem? Do we want to try and fix it?

Maybe wrap them in quotes?

lfdebrux:
> Maybe wrap them in quotes?

That could work 🤔 What do you think @NoraGDS?

NoraGDS:
maybe not quotes.

The capitalisation looks a little odd but may help the user to identify the individual skill/experience the error refers to, especially as the beginning of each error message is identical.

lfdebrux (approved):
Looks good to me!

PR #657, opened 2020-12-08
Change brief responses availability question to date type

lfdebrux:
Although this question asks for a simple date, it is a text question. It has been like this since it's initial commit to DOS1 [1]; I can find no evidence explaining why this question is not a date question.

To try and make it easier for users to enter a valid date this commit changes the question type to date for DOS5; this will not affect brief responses from previous frameworks and since DOS5 isn't live yet it shouldn't cause any issues with that framework (which is why I've called this a patch version change rather than a breaking change).

Hopefully this change will not be a problem going forward, although it will be interesting to see if anyone has been relying on this accepting input that is not a date.

[1]:
https://github.com/alphagov/digitalmarketplace-frameworks/commit/43edb2c4b64c93f33f34d6f5d33680ebb7357dc6#diff-1437d60cdf62bf2be98d263fe7b985e069f360262ee281719624776cf8618eb3

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
+++ b/frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
@@ -1,6 +1,6 @@
 name: Earliest start date
 question: When is the earliest {% if brief.lotSlug == "digital-outcomes" %}the team can start?{% elif brief.lotSlug == "digital-specialists" %}the specialist can start work?{% else %}you can recruit participants?{% endif %}
-hint: 'eg 31/12/2020'
+hint: "For example, 31 12 2020"
 question_advice: >
   The buyer needs
   {%- if brief.lotSlug == "digital-outcomes" %}
@@ -10,13 +10,12 @@ question_advice: >
   {%- else %}
   participants: {{ brief.researchDates }}
   {%- endif %}
-type: text
-smaller: true
+type: date
 validations:
-  -
-    name: answer_required
-    message: 'Enter a start date'
-  -
-    name: under_character_limit
-    message: "Your start date must be 100 characters or fewer"
+  - name: answer_required
+    message: "Enter a start date"
+
+  - name: invalid_format
+    message: "Start date must be a real date"
+
 empty_message: Set date
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
   "name": "digitalmarketplace-frameworks",
-  "version": "18.0.2",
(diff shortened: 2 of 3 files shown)


frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
```diff
@@ -10,13 +10,10 @@ question_advice: >
   {%- else %}
   participants: {{ brief.researchDates }}
   {%- endif %}
-type: text
-smaller: true
+type: date
```
bjgill, line 13: Looking at other DOS 5 questions with this type, is the hint still correct?

frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
```diff
@@ -10,13 +10,10 @@ question_advice: >
...
-  -
-    name: answer_required
-    message: 'You need to answer this question.'
-  -
-    name: under_character_limit
-    message: "Your start date must be no more than 100 characters."
+  - name: answer_required
+    message: Enter a start date
+  - name: invalid_format
+    message: Start date must be a real date
```
bjgill, line 18: Do we want these to be consistent with existing questions with type `date` - e.g. `frameworks/digital-outcomes-and-specialists-5/questions/briefs/awardedContractStartDate.yml`?

bjgill (commented):
Looks like a good improvement for users. Just a couple of questions about consistency with similar existing questions.

frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
lfdebrux: Yeah good idea

frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
lfdebrux: Oh good spot it's not

frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/availability.yml
lfdebrux: Hmm, actually looking at it closer I think the other questions need to have their error messages updated 🤔 They are currently not aligned with Design System guidance

This should happen some time before DOS5 is opened; if we don't have a ticket for it already I will create one now.

bjgill (approved):
:shipit:

PR #659, opened 2020-12-14
Remove headings from Day Rate question advice

domoscargin:
The headings made the question difficult to parse at a glance. Removing them places more focus on the
question and the input, rather than so much emphasis on the advice.

## Before
Screenshot 2020-12-14 at 10 11 09

## After
Screenshot 2020-12-14 at 11 41 12


lfdebrux:
I think we'll need to bump the version number

lfdebrux (approved):

PR #660, opened 2020-12-15
Fix bug in validation schema generator

lfdebrux:
Ticket: https://trello.com/c/LY1lCPUt/291-1-fix-schema-generator-in-frameworks-repo

Adding `{{ item }}` to dynamic list question error messages in #656 caused validation schema generation to fail because `item` was not present in the template context.

This PR adds `item` to the context for all questions; it is a bit of a hack but is the only fix I could think of.

Also, it was a bit worrying that we didn't notice that the schema generator was broken when we merged #656, so I've also added some functional tests of the schema generation scripts. I tested that this functional test fails as expected if the fix is not present.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,7 @@ jobs:
   unit_test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
 
@@ -19,7 +20,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: venv
-          key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements-dev.txt') }}
+          key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements*.txt') }}
 
       - name: Install dependencies
         run: make requirements-dev
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,9 +10,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install Ubuntu packages
-        run: sudo apt-get install libxml2-dev libxslt1-dev python-dev
-
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
@@ -44,9 +41,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-
-      - name: Install Ubuntu packages
-        run: sudo apt-get install libxml2-dev libxslt1-dev python-dev
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,3 +29,38 @@ jobs:
 
       - name: Run tests
         run: make test
+
+  functional_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6]
+        script:
+          - scripts/generate-assessment-schema.py g-cloud-12
+          - scripts/generate-assessment-schema.py digital-outcomes-and-specialists-5
+          - scripts/generate-search-config.py g-cloud-12 services --output-path=/tmp
+          - scripts/generate-search-config.py digital-outcomes-and-specialists-5 briefs --output-path=/tmp
+          - scripts/generate-validation-schemas.py --output-path=/tmp
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Ubuntu packages
+        run: sudo apt-get install libxml2-dev libxslt1-dev python-dev
+
+      - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -29,3 +29,38 @@ jobs:
...
+          - scripts/generate-assessment-schema.py digital-outcomes-and-specialists-5
+          - scripts/generate-search-config.py g-cloud-12 services --output-path=/tmp
+          - scripts/generate-search-config.py digital-outcomes-and-specialists-5 briefs --output-path=/tmp
+          - scripts/generate-validation-schemas.py --output-path=/tmp
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Ubuntu packages
+        run: sudo apt-get install libxml2-dev libxslt1-dev python-dev
```
bjgill, line 49: Do we need this? I don't think this repo uses `lxml`.

lfdebrux: Not sure, I just copied it from above...

lfdebrux: Here's one way to find out https://github.com/alphagov/digitalmarketplace-frameworks/pull/660/commits/8935b13b54b1b48a35aab5ca0994a4ef147e0946...

schema_generator/validation.py
```diff
@@ -124,7 +124,10 @@ def load_questions(schema_type, framework_slug, lot_slug):
     )
 
     manifest = loader.get_manifest(framework_slug, MANIFESTS[schema_type]['manifest']).filter(
-        {'lot': lot_slug},
+        {
+            "lot": lot_slug,
+            "item": "item",  # some dynamic list questions use `{{item}}` in template fields
+        },
```
bjgill, line 130: I'm not very familiar with content loader, so I'm having trouble understanding this. Is `DynamicList._make_dynamic_question#L453` (called from `DynamicList.filter`) not already adding `item` to the context?

lfdebrux: It does, but it is only called if `DynamicList.dynamic_field` is in the existing context (see https://github.com/alphagov/digitalmarketplace-content-loader/blob/master/dmcontent/questions.py#L335).

lfdebrux: But actually none of that code is reached anyway, because `dynamic=False` in the call to filter here.

.github/workflows/test.yml
lfdebrux: It worked! 😁

schema_generator/validation.py
bjgill: Ha - very true. So where will this `"item"` show up? I assume this won't be displayed to the user or appear in the API/logs?

lfdebrux: It never makes it into the schema, the error message displayed is treated as orthogonal to the validation check.

To be specific, in the current system, the API uses the schema to check a form response; if the validation fails the API generates a HTTP response containing the name of the validation check that failed; then the frontend looks at the response, cross-checks the name of the validation check with the content loader question, and looks up the error message for that validation failure.

bjgill: I think I now understand this. It looks as if validator only cares about the `name` field. And `item` only shows up in the `message` field.

So it's fine to give a placeholder because the validator doesn't use the field that this placeholder will appear in?

Might be nice to make this less confusing. But I don't know how one would do that without a fairly hefty amount of refactoring we probably don't have time for now. 

Could we perhaps be explicit that the only fields that use `item` are fields that aren't actually used by the validator?

lfdebrux: Assuming future changes hew to that design I do not expect the error messages to ever be present in the schema.

lfdebrux: > Could we perhaps be explicit that the only fields that use item are fields that aren't actually used by the validator?

Maybe I can write a test that asserts this...

schema_generator/validation.py
```diff
@@ -124,7 +124,10 @@ def load_questions(schema_type, framework_slug, lot_slug):
     )
 
     manifest = loader.get_manifest(framework_slug, MANIFESTS[schema_type]['manifest']).filter(
-        {'lot': lot_slug},
+        {
+            "lot": lot_slug,
+            "item": "item",  # some dynamic list questions use `{{item}}` in template fields
```
bjgill, line 129:
```suggestion
            # Some dynamic list questions use `{{item}}` in template fields, so we need to provide it in the context.
            # Should not appear in the generated schema.
            "item": "item",
```

schema_generator/validation.py
bjgill: That's a good idea. I've made a suggestion on how we could change the wording to make it clearer for developers who don't already know how all the schema generation works.

lfdebrux:
I've amended the commit to add tests, and also to move the point where `item` is added to the context into a smaller scope (this was also needed to make tests work). It also adds more details to the comment. How does this look @bjgill?

bjgill (approved):
:tada:

lfdebrux:
GitHub actions was failing because it thought Python 3.9 wasn't present (see https://github.com/alphagov/digitalmarketplace-frameworks/pull/660/checks?check_run_id=1557392303), invalidating the cache following @gidsg 's suggestion seems to have fixed it.

PR #662, opened 2020-12-21
add contract title to descriptions

gidsg:
https://trello.com/c/75hbcRW8/607-3-implement-dos-5-e-signature

We are using this value in a few different places so this adds it to frameworks so it can be referenced from multiple repos.


gidsg:
@lfdebrux I've moved this to a separate file for reasons explained in the commit message. Please can you confirm you're happy with this approach?

lfdebrux:
> @lfdebrux I've moved this to a separate file for reasons explained in the commit message. Please can you confirm you're happy with this approach?

Fine by me. It might be worth explaining in a little more detail why the change is needed; for instance I don't really understand why the key needs to be in the schema. But otherwise this solution seems fine to me.

gidsg:
@lfdebrux it appears the test walk and validate every file in the directory so we need to update the schema to prevent the tests from failing.

PR #664, opened 2021-01-05
Update DOS5 URLs

gidsg:
https://trello.com/c/75hbcRW8/607-2-implement-dos-5-e-signature
N.B. There is no call off contract URL for this framework. I think this is not linked to so _should_ be OK.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-5/messages/urls.yml
+++ b/frameworks/digital-outcomes-and-specialists-5/messages/urls.yml
@@ -1,4 +1,3 @@
-call_off_contract_url: https://www.gov.uk/government/publications/digital-outcomes-and-specialists-5-call-off-contract/__placeholder__
-framework_agreement_pdf_url: https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/738942/digital-outcomes-and-specialists-5-framework-agreement.pdf/__placeholder__
-framework_agreement_url: https://www.gov.uk/government/publications/digital-outcomes-and-specialists-5-framework-agreement/__placeholder__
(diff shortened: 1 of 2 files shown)


frameworks/digital-outcomes-and-specialists-5/messages/urls.yml
```diff
@@ -1,4 +1,3 @@
-call_off_contract_url: https://www.gov.uk/government/publications/digital-outcomes-and-specialists-5-call-off-contract/__placeholder__
-framework_agreement_pdf_url: https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/738942/digital-outcomes-and-specialists-5-framework-agreement.pdf/__placeholder__
-framework_agreement_url: https://www.gov.uk/government/publications/digital-outcomes-and-specialists-5-framework-agreement/__placeholder__
-supplier_guide_url: https://www.gov.uk/guidance/digital-outcomes-and-specialists-suppliers-guide#how-to-apply-to-the-framework/__placeholder__
+framework_agreement_pdf_url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/947688/DOS5-framework-award-form.odt
```
bjgill, line 1: > pdf

> .odt

Is this OK?

bjgill (approved):
Looks sensible. Just one query. But if that's fine, feel free to merge

frameworks/digital-outcomes-and-specialists-5/messages/urls.yml
gidsg: @bjgill We are only making this available in ODT currently, I would rather not rename the key as it will break things.

PR #665, opened 2021-01-11
Add template logic to 'How evaluated' description

gidsg:
https://trello.com/c/YZJhCqq2/238-3-dos5-public-brief-page-add-mandatory-assessment-method-to-summary-text

Rather than hardcoding this logic in the app, as previously in: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/37f0d7da87b41843f809c107b7cd376d6487308f/app/main/helpers/brief_helpers.py#L33-L45 we want to move this to the frameworks repository.

### DIFF
--- a/frameworks/digital-outcomes-and-specialists-5/manifests/display_brief.yml
+++ b/frameworks/digital-outcomes-and-specialists-5/manifests/display_brief.yml
@@ -59,6 +59,7 @@
     - niceToHaveRequirementsParticipants
 -
   name: How suppliers will be evaluated
+  summary_page_description: "All suppliers will be asked to provide a {% if lotSlug == 'digital-specialists' %}work history{% else %}written proposal{% endif %}."
   questions:
     - numberOfSuppliersOutcomes
     - numberOfSuppliersSpecialists
--- a/frameworks/digital-outcomes-and-specialists-4/manifests/display_brief.yml
+++ b/frameworks/digital-outcomes-and-specialists-4/manifests/display_brief.yml
@@ -59,6 +59,7 @@
     - niceToHaveRequirementsParticipants
 -
   name: How suppliers will be evaluated
+  summary_page_description: "All suppliers will be asked to provide a {% if lotSlug == 'digital-specialists' %}work history{% else %}written proposal{% endif %}."
   questions:
     - numberOfSuppliersOutcomes
     - numberOfSuppliersSpecialists
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "digitalmarketplace-frameworks",
-  "version": "18.1.6",
+  "version": "18.1.7",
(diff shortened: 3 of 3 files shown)


lfdebrux:
Hmm, this looks like it should work, but it is hard to test without the corresponding brief-responses frontend change, is that available to check out?

frameworks/digital-outcomes-and-specialists-5/manifests/display_brief.yml
```diff
@@ -59,6 +59,7 @@
     - niceToHaveRequirementsParticipants
 -
   name: How suppliers will be evaluated
+  summary_page_description: "All suppliers will be asked to provide a {% if lotSlug == 'digital-specialists' %}work history{% else %}written proposal{% endif %}."
```
lfdebrux, line 62: I've been testing in flask shell, I think it should be `lot` instead of `lotSlug`

```suggestion
  summary_page_description: "All suppliers will be asked to provide a {% if lot == 'digital-specialists' %}work history{% else %}written proposal{% endif %}."
```

lfdebrux:
I've been testing in flask shell and can't get this to work 🤔

frameworks/digital-outcomes-and-specialists-5/manifests/display_brief.yml
lfdebrux, line 62: When testing in flask shell I get the error 

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/laurencedebruxelles/Code/digitalmarketplace/digitalmarketplace-content-loader/dmcontent/content_loader.py", line 192, in __getattribute__
    return field.render(context)
  File "/Users/laurencedebruxelles/Code/digitalmarketplace/digitalmarketplace-content-loader/dmcontent/utils.py", line 66, in render
    raise ContentTemplateError(e.message)
dmcontent.errors.ContentTemplateError: 'lotSlug' is undefined
```

I think it should be `brief['lotSlug']`, as the manifest gets filtered by `brief` in the buyer frontend before display

https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/a54b9a071525ddcacac071973d3be65ed96c2b0d/app/main/views/marketplace.py#L183

gidsg:
@lfdebrux Hmm, this does appear to work for me, as per your suggestion I've created a draft PR so the tests can run against it https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1128
This logic is tested by https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/21403ced3441550d7908577af02bf19ba7ae3f9b/tests/main/views/test_marketplace.py#L594-L619

lfdebrux:
@gidsg Ah, brilliant! I can see what I did wrong now also, got it working for me so going to approve. Nice one!

lfdebrux (approved):

PR #671, opened 2021-02-03
Remove private: true from package.json

lfdebrux:
We want to publish this package on NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,5 @@
   "version": "18.1.9",
   "description": "Data files for Digital Marketplace’s procurement frameworks",
   "repository": "git@github.com:alphagov/digitalmarketplace-frameworks",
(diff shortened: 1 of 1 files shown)


bjgill (approved):

package.json
```diff
@@ -4,5 +4,4 @@
   "description": "Data files for Digital Marketplace’s procurement frameworks",
   "repository": "git@github.com:alphagov/digitalmarketplace-frameworks",
   "author": "enquiries@digitalmarketplace.service.gov.uk",
```
bjgill, line 6: I assume NPM's JSON parser is happy with this trailing `,`?

lfdebrux: It is not 😢

bjgill (approved):

PR #674, opened 2021-03-04
Update lot questions descriptions

domoscargin:
https://trello.com/c/r8CxfKTO/820-5-replace-choose-a-lot-page-in-create-a-service-journey

We're adjusting the journey to create a service, and we're taking the opportunity to update the content.


lfdebrux:
These are from CCS and our content designer I assume?

lfdebrux:
I wonder if there's any point making these changes to these frameworks, seeing as they are now closed for applications and the content changes will never show up... what if instead we created DOS6 and G13 now?

@bjgill what do you think?

domoscargin:
> These are from CCS and our content designer I assume?

These are content designer changes, though I'm not sure if we need to run it through somebody else on the CCS side?

bjgill:
Laurence - if the changes are made here, they'll appear in G13/DOS6 when [we create them by copying G12/DOS5](https://alphagov.github.io/digitalmarketplace-manual/content-and-frameworks/adding-frameworks.html#make-the-content). So I think it's unlikely to matter where we make the changes.

The only question I can think of is: does this change anything that the functional tests rely on/test?

lfdebrux:
> Laurence - if the changes are made here, they'll appear in G13/DOS6 when [we create them by copying G12/DOS5](https://alphagov.github.io/digitalmarketplace-manual/content-and-frameworks/adding-frameworks.html#make-the-content). So I think it's unlikely to matter where we make the changes.

True, it's just a question of preference.

> The only question I can think of is: does this change anything that the functional tests rely on/test?

Almost certainly not.

lfdebrux (approved):

lfdebrux (approved):

PR #678, opened 2021-03-19
Use bulleted lists for G-Cloud 12 questions

alex9smith:
We spotted this bug while working on content loader changes for these questions[[1]].

[1]: https://github.com/alphagov/digitalmarketplace-content-loader/pull/140


bjgill (approved):
Looks sensible. What was the bug - why was this not previously a problem?

alex9smith:
Without the class `govuk-list--bullet` the list items don't have bullet points next to each of them which looks a bit strange:

image
(This is a screenshot from the other work I'm doing so it might be slightly different to production)

I'm not sure why this wasn't a problem previously. Either it was a problem and we missed it while checking, or we've lost some custom behaviour that added this class at some point in migrating from the DMp frontend components to the GOV.UK ones. I suspect the latter is more likely. It doesn't affect DOS5 because the same questions on DOS5 don't have hardcoded HTML.

PR #687, opened 2021-06-14
Use 'contracted out service' in question title

alex9smith:
This was updated everywhere else based on feedback from the Tax Centre of Excellence, but we missed it in the question title.


gidsg (approved):
👍 🆙

